PR #5, opened 2020-09-14
Hyrax-doi integration

cjcolvar:
Bring `hyrax-doi` in and enable multi-tenancy by creating a DataCiteEndpoint for storing per-tenant configuration.  This is configurable then from the account edit page and switched just like fedora/solr/redis.

Overrides in this PR: account edit view and some methods of account and accounts controller (in the engine).

Also included in this PR is switching the internal test hyku application to a special fork which includes some unmerged commits and leaves the user elevation feature out.  We'll want to update to latest in the default branch eventually but this is good for now.


app/models/datacite_endpoint.rb
```diff
@@ -0,0 +1,29 @@
...
+  def switch!
+    Hyrax::DOI::DataCiteRegistrar.mode = mode
+    Hyrax::DOI::DataCiteRegistrar.prefix = prefix
+    Hyrax::DOI::DataCiteRegistrar.username = username
+    Hyrax::DOI::DataCiteRegistrar.password = password
+  end
+
+  # Remove the solr collection then destroy this record
+  def remove!
+    destroy
```
BertZZ, line 15: What destroy is this calling? This method seems to be missing a test.

mankind: @BertZZ  it is calling activerecord destroy.
https://apidock.com/rails/ActiveRecord/Relation/destroy

cjcolvar: I was lazy and wasn't really writing tests beyond what hyku was already testing.  I'm not entirely sure if this method is being used but left it in the copy from solr endpoint.  (Looks like I need to fix that comment at the least.)  I'll do a quick look and see if I can remove it or add a test if it is still necessary.

BertZZ: @mankind That method takes an id usually which is what thew me here.

mankind: @BertZZ not always. You can fetch the object and just call destroy on it with no arguments.

cjcolvar: I don't think the method is actually called for DataCiteEndpoint but I left it in to match the other endpoints.  Tests added including testing that the destroy cascades from the account destroy since I didn't override CleanupAccountJob to call `remove!` explicitly for DataCiteEndpoint.

BertZZ (approved):
:+1:

PR #6, opened 2020-09-16
Docker setup for development and testing environments

cjcolvar:
Drawing from:
- Avalon's multistage Dockerfile: https://github.com/avalonmediasystem/avalon/blob/master/Dockerfile
- Hyku's docker-compose.yml: https://github.com/samvera/hyku/blob/master/docker-compose.yml
- https://dev.to/bgord/multiple-postgres-databases-in-a-single-docker-container-417l

Also switched internal test hyku to use temporary hyku fork that has specific commits from hyku master and PRs but without the user elevation feature: https://github.com/ubiquitypress/hyku/tree/hyku_2/3_without_user_elevation


cjcolvar:
Closing in favor of #9.

PR #8, opened 2020-09-21
Create DataCiteEndpoint when other endpoints are created (mostly)

cjcolvar:
When I tried to deploy this to AH Demo it couldn't load the app and I think it couldn't even run the migrations because datacite_endpoint didn't exist in the DB yet.  I think this was maybe a chicken/egg problem.  This change should help with that and make it more consistent with the other endpoints.  Testing setup was also improved in the process.


BertZZ (approved):
:+1:

PR #10, opened 2020-09-23
Refactor hyrax-doi integration

cjcolvar:
- Add hyku_addons:install generator
- Mark specific generators as app-only instead of operating on the engine
- Force `eager_load` in development environment to deal with autoloading issue
- Switch rails `default_url_options` so url the DOI resolves to is correct


cjcolvar:
I'm not concerned about the coverage failing since the only lines not covered are only meant for development environments.

BertZZ (approved):
:+1:

PR #13, opened 2020-10-19
UI for settings key that hold hash value

mankind:


cjcolvar (approved):
:+1: Looks good.  Thanks @mankind!

PR #16, opened 2020-12-08
Further settings work

BertZZ:
Neatens up the settings page a little to make it more user friendly. 
Adds validations for the fields that are likely to stay on this form. 
Tests to follow.

### DIFF
--- a/app/models/concerns/hyku_addons/account_behavior.rb
+++ b/app/models/concerns/hyku_addons/account_behavior.rb
@@ -36,6 +36,7 @@ module HykuAddons
     end
 
     private
+    
       def validate_email_format
         return unless settings['email_format'].present?
         settings['email_format'] = settings['email_format'].grep(/@\S*\.\S*/)
--- a/app/models/concerns/hyku_addons/account_behavior.rb
+++ b/app/models/concerns/hyku_addons/account_behavior.rb
@@ -36,7 +36,7 @@ module HykuAddons
     end
 
     private
-    
+
       def validate_email_format
         return unless settings['email_format'].present?
         settings['email_format'] = settings['email_format'].grep(/@\S*\.\S*/)
--- a/app/models/concerns/hyku_addons/account_behavior.rb
+++ b/app/models/concerns/hyku_addons/account_behavior.rb
@@ -24,7 +24,7 @@ module HykuAddons
       after_initialize :set_jsonb_help_texts_default_keys, :set_jsonb_work_unwanted_fields_default_keys
       after_initialize :set_jsonb_required_json_property_default_keys, :set_jsonb_html_required_default_keys
       after_initialize :set_jsonb_metadata_labels_default_keys, :set_jsonb_licence_list_default_keys
-      before_validation :map_array_fields, on: [:create, :update_single, :update, :edit]
+      before_validation :map_array_fields
       before_save :remove_settings_hash_key_with_nil_value
       validates :gtm_id, :demo_gtm_id, format: { with: /GTM-[A-Z0-9]{4,7}/, message: "Invalid GTM ID" }, allow_blank: true
       validates :contact_email, format: { with: email_validation_format }, allow_blank: true
--- a/app/models/concerns/hyku_addons/account_behavior.rb
+++ b/app/models/concerns/hyku_addons/account_behavior.rb
@@ -3,7 +3,9 @@
 module HykuAddons
   module AccountBehavior
     extend ActiveSupport::Concern
+    include ActiveModel::Validations
     included do
+      email_validation_format = URI::MailTo::EMAIL_REGEXP
       belongs_to :datacite_endpoint, dependent: :delete
       has_many :children, class_name: "Account", foreign_key: "parent_id", dependent: :destroy, inverse_of: :parent
       belongs_to :parent, class_name: "Account", inverse_of: :parent, foreign_key: "parent_id", optional: true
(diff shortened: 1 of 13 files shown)


app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -83,5 +102,12 @@ def remove_settings_hash_key_with_nil_value
           settings[key].delete_if { |_hash_key, value| value.blank? } if settings[key].class == Hash
         end
       end
+
+      def map_array_fields
```
cjcolvar, line 106: This seems more like a controller or form concern to me than a model concern.  What do you think?

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -31,6 +37,19 @@ def datacite_endpoint
 
     private
 
+      def validate_email_format
```
cjcolvar, line 40: It's not clear to me if this method is intended for checking if `email_format` is in the proper format or is intended to do data manipulation.  Same for `validate_contact_emails`.  I would expect validation methods to only be checking for proper format and data manipulation to be happening in the controller or form or setter methods.

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -3,7 +3,9 @@
 module HykuAddons
   module AccountBehavior
     extend ActiveSupport::Concern
+    include ActiveModel::Validations
     included do
+      email_validation_format = URI::MailTo::EMAIL_REGEXP
```
cjcolvar, line 8: Do we need to store this in a variable?  It feels like it should either be a constant or skipped and use `URI::MailTo::EMAIL_REGEXP` directly in the `validates` below.

cjcolvar (commented):
Looks good. I just had a few questions/suggestions.

app/models/concerns/hyku_addons/account_behavior.rb
okeen: There is a lot of validation logic into the Account model that is not intrinsic to the Account itself, but dependent on the context (for example, should an account always validate the settings email format, or rather only do it when you are updating settings?). 

It looks like we could extract all the settings functionality into a Form object as Chris says. https://codeclimate.com/blog/7-ways-to-decompose-fat-activerecord-models/

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -22,7 +24,11 @@ module AccountBehavior
       after_initialize :set_jsonb_help_texts_default_keys, :set_jsonb_work_unwanted_fields_default_keys
       after_initialize :set_jsonb_required_json_property_default_keys, :set_jsonb_html_required_default_keys
       after_initialize :set_jsonb_metadata_labels_default_keys, :set_jsonb_licence_list_default_keys
+      before_validation :map_array_fields
       before_save :remove_settings_hash_key_with_nil_value
+      validates :gtm_id, :demo_gtm_id, format: { with: /GTM-[A-Z0-9]{4,7}/, message: "Invalid GTM ID" }, allow_blank: true
```
okeen, line 29: I think putting this into Rails' Validator clases would make it easier to read, maintain and refactor in the future:

`validates :gtm_id, :demo_gtm_id, gtm: true`

https://api.rubyonrails.org/classes/ActiveModel/Validator.html

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -31,6 +37,19 @@ def datacite_endpoint
...
     private
 
+      def validate_email_format
+        return unless settings['email_format'].present?
+        settings['email_format'] = settings['email_format'].grep(/@\S*\.\S*/)
+      end
+
+      def validate_contact_emails
+        email_validation_format = URI::MailTo::EMAIL_REGEXP
+        ['weekly_email_list', 'monthly_email_list', 'yearly_email_list'].each do |key|
```
okeen, line 47: Should we add a class property + accessor for the weekly, monthly and yearly list array?

@@AVAILABLE_EMAIL_PERIODS = %w(week month year).map{ |period| "#{period}ly_email_list" }
cattr_reader :AVAILABLE_EMAIL_PERIODS

We could even set the periods on the settings.yml file

app/views/hyku_addons/account_settings/index.html.erb
```diff
@@ -19,36 +19,34 @@
 
     <% ['contact_email', 'gtm_id', 'demo_gtm_id', 'live', 'institutional_relationship', 'creator_fields', 'contributor_fields'].each do |fname| %>
```
okeen, line 20: We could extract this into a rails helper that accepts a block. That would improve readability which is extremely important on the views IMO.

Something like:
```

def for_available_setting_field
  Account.contact_settings_fields.each do |field|
    capture(yield(field))
  end
end
```

And then use it as:

```
<% for_available_setting_field do |field | %>
....

```

app/views/hyku_addons/account_settings/index.html.erb
```diff
@@ -19,36 +19,34 @@
...
       <tr>
-        <td id= "link-<%= fname %>"><b><%= fname %>:</b> <%= @account && @account.settings && @account.settings[fname] %> </td> <br>
-        <td><%= link_to "Edit #{fname}", hyku_addons.edit_admin_account_setting_path(id: @account, partial_name: 'render_text_fields', field_name: fname), class: "btn btn-xs btn-default",  remote: true %></td>
+        <td id= "link-<%= fname %>"><b><%= fname.humanize %>:</b> <%= @account && @account.settings && @account.settings[fname] %> </td> <br>
```
okeen, line 22: I think we can extract this into a helper method + presenters.
Something like:

<td><%= link_to_edit_account_settings %></td>

https://www.rubyguides.com/2019/09/rails-patterns-presenter-service/

okeen (commented):
I have attached a few comments inline. Sorry if the formatting is not the best :(

My very first GH review ever :)

app/models/concerns/hyku_addons/account_behavior.rb
BertZZ: Yeah you are probably right here.

app/models/concerns/hyku_addons/account_behavior.rb
BertZZ: I see what you are saying @cjcolvar let me try something else

app/models/concerns/hyku_addons/account_behavior.rb
BertZZ: Yeah let me try moving it to the controller

app/views/hyku_addons/account_settings/index.html.erb
BertZZ: I agree but I'm not sure it's something I want to spend time on now

cjcolvar (approved):
Looks good! :tada:

cjcolvar:
I'm not really sure what's up with the total coverage dropping when the diff coverage is 100%.

PR #17, opened 2020-12-09
Generic work JSON fields

cjcolvar:


cjcolvar:
I know the diff coverage is below the threshold and the total coverage dropped as a result, but I'm going to go ahead and merge this since it needs to get to testing.  Refactoring should help clean this up and deal with coverage gaps.

PR #25, opened 2020-12-18
Super admin settings

cjcolvar:
This does the initial work to setup the ability to save settings from the super admin proprietor account edit page.


cjcolvar:
The only failed test was the feature spec which is intermittently failing at the moment and is unrelated to this PR so I'm going to go ahead and merge.

PR #30, opened 2020-12-28
Add RIS export option for Works

cjcolvar:
- Rebase and squash of #29
- Use polymorphic_path for link


prwhitehead (dismissed):
Thanks for fixing up the route issue.

prwhitehead:
@cjcolvar could you look over this PR when you get back after the New Year please :)

cjcolvar (commented):
@prwhitehead This all looks great!
I was thinking it might make more sense for the `export_as_ris` method implementation to move into `HykuAddons::RisContentNegotiation` and then have the presenter `delegate :export_as_ris, to: :solr_document`.  What do you think?

cjcolvar:
I ended up moving the spec test for the mime type to a test on the engine to make rubocop happy about having the lib spec test describe a class or module.

prwhitehead:
> @prwhitehead This all looks great!
> I was thinking it might make more sense for the `export_as_ris` method implementation to move into `HykuAddons::RisContentNegotiation` and then have the presenter `delegate :export_as_ris, to: :solr_document`. What do you think?

prwhitehead:
@cjcolvar Thank you for the comments. I've moved the `export_as_ris` to the Negotiator as you suggested.

cjcolvar:
I'm going to go ahead and merge this since the failing tests aren't related to this change.  I'm trying to fix the failing feature tests by switching to apparition as the capybara JS driver: #36

PR #31, opened 2020-12-28
More work types

cjcolvar:
Ran hyrax:work generator (with --skip-namespaces option)
Modified model, form, and presenters
Registered the work type in an initializer


spec/spec_helper.rb
```diff
@@ -40,7 +40,7 @@
 
   # This option will default to `:apply_to_host_groups` in RSpec 4 (and will
   # have no way to turn it off -- the option exists only for backwards
-  # compatibility in RSpec 3). It causes shared context metadata to be
+  # compatibility in RSpec 3). It causes helpers context metadata to be
```
cjcolvar, line 43: This looks like a search/replace that shouldn't have happened.

prwhitehead: Thats a weird change and I dont remember making it or doing any large search/replace, so I've undone the change. Sorry about that.

prwhitehead: Wierdly I'm not seeing this in the repo: 

https://github.com/ubiquitypress/hyku_addons/blob/Feature/RisDocumentType-chris/spec/spec_helper.rb#L43

Seems to have been changed. Not sure why this is shown as different from what was commited.

cjcolvar: Looks like that link is for the branch for #30.  I'm seeing it in the `more_work_types` branch: https://github.com/ubiquitypress/hyku_addons/blob/more_work_types/spec/spec_helper.rb#L43

prwhitehead: Strange. i'd not merged that branch. :-/

app/forms/hyrax/book_contribution_form.rb
```diff
@@ -11,19 +12,16 @@ class BookContributionForm < Hyrax::Forms::WorkForm
...
     self.model_class = ::BookContribution
     self.terms = %i[title resource_type creator alt_title contributor rendering_ids abstract date_published
-                      institution org_unit project_name funder fndr_project_ref series_name book_title editor volume edition  pagination
-                      publisher place_of_publication isbn issn eissn date_accepted date_submitted official_link related_url
-                      language license rights_statement rights_holder doi alternate_identifier related_identifier refereed
-                      keyword dewey library_of_congress_classification add_info]
+                    institution org_unit project_name funder fndr_project_ref series_name book_title editor volume edition
+                    publisher place_of_publication isbn issn eissn date_accepted date_submitted official_link related_url
+                    language license rights_statement rights_holder doi alternate_identifier related_identifier refereed
+                    keyword dewey library_of_congress_classification add_info pagination]
```
cjcolvar, line 18: I'm not sure if this is a safe move or not because I think the order in the terms array sets the order of the fields on the deposit form.

app/forms/concerns/hyku_addons/work_form.rb
```diff
@@ -14,8 +14,10 @@ def contributor_list
 
     class_methods do
       def common_fields
-        [:title, :resource_type, :alternative_name, :project_name, :institution, :abstract, creator_fields, contributor_fields,
-         date_published_fields, funder_fields]
+        [:title, :resource_type, :alternative_name, :project_name, :institution, :abstract, :official_link, :related_url,
+         :language, :license, :rights_statement, :rights_holder, :doi, :peer_reviewed, :keywords, :dewey, :library_of_congress_classification,
```
cjcolvar, line 18: I think `:peer_reviewed` is called `:refereed` in the model.  According to the spreadsheet it isn't common to all work types so we can drop it from this list.

PR #34, opened 2021-01-05
Adds code for GTM support

BertZZ:


cjcolvar (commented):
This is looking good.  I have two questions/thoughts:
1. I think/hope it should be possible to use `content_for` blocks in a new partial to inject the GTM code without having to override the hyrax layout.  Looks like `:head` and `:precontainer_content` might be the best hooks.
1.  Do we need to support multiple GTM ids at this point? (Or am I getting it confused with Google analytics ids?)

cjcolvar (approved):
Looks great!

cjcolvar (approved):

PR #36, opened 2021-01-05
Test using apparition instead of webdrivers

cjcolvar:


cjcolvar:
I'm giving up on this for now as it doesn't magically fix the issues we're having with feature tests.

PR #37, opened 2021-01-06
Fix Issues with Creator Deposit Form

prwhitehead:
There were a few issues with the form that should be fixed by this. 

I have stopped short of refactoring the JS into a class, as it felt like overkill - wrapping procedural jQuery event listeners in a class method without actually refactoring to properly encapsulate the object. Instead I have tried to make the JS (and to a lesser extend the HTML) more readable and less inconsistent. 

I have tried to avoid breaking anything, but there seems to be a block of code without any matching HTML reference, so for now that has been commented out.


cjcolvar (approved):
I skimmed this quickly and looks good enough to me.  This same work will need to be done on the contributor and editor fields I believe.  Refactoring to DRY up those three cases might make sense to do now.

PR #39, opened 2021-01-07
Hides sign up link based on setting

BertZZ:
Adds functionality for toggle of sign up to a repository


app/controllers/concerns/hyku_addons/registrations_controller.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  class RegistrationsController < Devise::RegistrationsController
```
cjcolvar, line 4: Is this meant to be a module that gets included into `Devise::RegistrationsController` or a class that replaces it?  If a class, then it should be moved out of the `concerns` directory.  I think this could work really well as a module if the new and create overrides were removed in favor of a `before_action` which does the redirect or raises an error which gets handled in ApplicationController (by performing the redirect).  Probably the latter of those two would be the cleanest.  I can point to some examples if you'd like.

BertZZ: Please :)

cjcolvar: Here's an example of the handling of the raised error.  Looks like it could live in the module instead of ApplicationController.
https://github.com/samvera/hydra-head/blob/master/hydra-core/app/controllers/concerns/hydra/controller/controller_behavior.rb#L11-L32

cjcolvar:
@BertZZ I did a little bit of tinkering but there are still failing tests.

app/views/hyku_addons/account_settings/settings/_boolean_field_content.html.erb
```diff
@@ -2,7 +2,7 @@
 <%=f.fields_for :settings do |ff| %>
   <%= ff.label field_name %><br>
   <%= ff.label t("settings.helptexts.#{field_name}") %><br>
-  <%= ff.check_box field_name, {checked: @account.settings[field_name]}, true, false %>
```
cjcolvar, line 5: Why did you decide to remove this?  It looks like this was setting the field in the form based upon the saved value, but now it looks like it will always show as unchecked even when it has be saved as true.

spec/features/sign_up_settings_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it "allows the user to create an account with a valid email format" do
+      visit '/users/sign_up'
+      fill_in 'user_display_name', with: "Test User"
+      fill_in 'user_email', with: "test@test.com"
+      fill_in 'user_password', with: "Potato123!"
+      fill_in 'user_password_confirmation', with: "Potato123!"
+      expect { click_on 'Create account' }.to change { User.count }.by(1)
+    end
+
+    it "prevents users from createing an account containing the correct email formats" do
```
cjcolvar, line 22:
```suggestion
    it "prevents users from creating an account without the correct email formats" do
```

app/views/hyku_addons/account_settings/settings/_boolean_field_content.html.erb
```diff
@@ -2,7 +2,7 @@
...
+  <%= ff.check_box field_name, {}, "true", "false" %>
```
cjcolvar, line 5: I see you switched these values from booleans to Strings.  This appears to break the following test.  I'm curious why you made this change and whether they should get switched back to booleans or the test updated to reflect the type change.

>  1) HykuAddons::AccountBehavior Settings Customisations settings from environment variable check all boolean truthy values
     Failure/Error: expect(account.settings[key]).to be_a(TrueClass)
       expected "true" to be a kind of TrueClass
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.10.1/lib/rspec/support.rb:97:in `block in <module:Support>'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-support-3.10.1/lib/rspec/support.rb:106:in `notify_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/fail_with.rb:35:in `fail_with'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/handler.rb:40:in `handle_failure'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/handler.rb:56:in `block in handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/handler.rb:27:in `with_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/handler.rb:48:in `handle_matcher'
     # ./vendor/bundle/ruby/2.7.0/gems/rspec-expectations-3.10.1/lib/rspec/expectations/expectation_target.rb:65:in `to'
     # ./spec/models/concerns/hyku_addons/account_behavior_spec.rb:131:in `block (5 levels) in <top (required)>'
     # ./spec/models/concerns/hyku_addons/account_behavior_spec.rb:129:in `each'
     # ./spec/models/concerns/hyku_addons/account_behavior_spec.rb:129:in `block (4 levels) in <top (required)>'

cjcolvar (commented):
This looks like it is almost there.  I'm only seeing one failing test related to this code (the other one we can ignore for now since it is a known flaky test).

app/views/hyku_addons/account_settings/settings/_boolean_field_content.html.erb
BertZZ: This might be an error

app/views/hyku_addons/account_settings/settings/_boolean_field_content.html.erb
BertZZ: I made this change because the first version wasn't saving as false, it might have been due to the line I removed above?

BertZZ: Ill look into it tomorrow morning

app/views/hyku_addons/account_settings/settings/_boolean_field_content.html.erb
BertZZ: When these are set as Boolean's they are still saved as strings when I look at them in the rails console. I have kept them as Strings and will change the test

cjcolvar (approved):
Looks good!

PR #42, opened 2021-01-08
Simple rake task to create an account

cjcolvar:
This gets run as `bundle exec rake hyku:account:create[pacific]`

### DIFF
--- a/lib/tasks/account.rake
+++ b/lib/tasks/account.rake
@@ -1,0 +1,10 @@
+# frozen_string_literal: true
+
+namespace :hyku do
+  namespace :account do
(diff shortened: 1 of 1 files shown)


fradeve:
Hey @cjcolvar that's great, thanks... can I pass the name as an argument to the command? If that is the case, how do I do it?

cjcolvar:
The name goes inside the square brackets:
`bundle exec rake hyku:account:create[name-of-new-tenant]`

fradeve:
Sorry I should have read the first comment, my bad!
It can be merged, IMHO :+1:

PR #45, opened 2021-01-13
Class fields are too new for Uglifier

cjcolvar:
I was seeing errors when trying to precompile assets for production environment (`RAILS_ENV=production bundle exec rake app:assets:precompile` in the docker container).
>Uglifier::Error: Unexpected token: operator (=)
--
 87160     funder_name.css('border-color', '#ccc');
 87161   }
 87162 }
 87163 ;
 87164 // Example:
 87165 // $("body").trigger("clear_inputs", [$(el)])
 87166 
 87167 class InputClearableListener {
    =>   eventName = "clear_inputs"
 87169 
 87170   constructor(){
 87171     this.registerListeners()
 87172   }
 87173 
 87174   registerListeners(){
 87175     $("body").on(this.eventName, this.onEvent.bind(this))
 87176   }

It looks like these compile if moved into the constructor.

### DIFF
--- a/app/assets/javascripts/hyku_addons/change_toggleable_listener.js
+++ b/app/assets/javascripts/hyku_addons/change_toggleable_listener.js
@@ -15,15 +15,15 @@
 //  </div>
 
 class ChangeToggleableListener {
-  parentSelector = "[data-toggleable]"
-  groupAttributeName = "data-toggleable-group"
-  groupSelector = `[${this.groupAttributeName}]`
-  controlSelector = "[data-toggleable-control]"
-  requiredSelector = "[data-required]"
-  eventName = "toggleable_group"
-  afterHiddenAttributeName = "data-after-toggleable-hidden"
-
   constructor(){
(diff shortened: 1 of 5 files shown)


prdanelli:
Could we update Uglifier, or is that brought in from upstream?

I suppose an alternative would be move the code outside of the asset pipeline, and require it from a public folder.

cjcolvar:
I don't think we can upgrade Uglifier to handle this.  I was using an online ESLint and none of the available ECMA Version options liked the class fields.  I don't think we should bypass Uglifier because I'd be afraid of not having browser support for it. (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields#browser_compatibility)

I think this should spawn a discussion about browser support and which versions we are claiming to support for which parts of the application (backend vs. frontend).

prdanelli:
Thanks for explaining @cjcolvar. I'll keep in mind that the class properties are not yet available to us.

prdanelli (approved):

PR #47, opened 2021-01-14
Hotfix/form js fixes

prdanelli:
Just a couple of fixes to errors I spotted this morning.


okeen (approved):
Looking good mate and thanks for teaching me about Array.filter(klazz) !

PR #49, opened 2021-01-14
AH-132 - Flexible Work Type Readers

prdanelli:
The idea of this PR is to make work type readers more flexible, so that each type can potentially have their own.


cjcolvar (commented):
This all looks good to me so far.  I'm not quite sure yet what next steps should be with respect to `hyrax-doi`.  Feel free to keep running with this.  It would be good to have a test for crosswalking to DataCite DOI XML like https://github.com/samvera-labs/hyrax-doi/blob/main/spec/services/bolognese/readers/hyrax_work_reader_spec.rb#L62-L129

prdanelli:
I realised that we were missing a bunch of keys from the RIS export that the legacy repositories had: 

https://gitlab.com/ubiquitypress/hyku/-/blob/test/app/controllers/concerns/ubiquity/citations_export_concern.rb

These have been incorporated, but its ugly, ugly code and needs to be refactored now that I have passing specs for the structure.

lib/hyku_addons/engine.rb
```diff
@@ -188,14 +188,13 @@ def self.dynamically_include_mixins
...
-      Bolognese::Metadata.prepend Bolognese::Readers::UbiquityGenericWorkReader
-      Hyrax::GenericWorksController.prepend HykuAddons::WorksControllerAdditionalMimeTypesBehavior
       User.include HykuAddons::UserEmailFormat
+      Bolognese::Writers::RisWriter.include Bolognese::Writers::RisWriterBehavior
     end
 
     # Use #to_prepare because it reloads where after_initialize only runs once
     # This might slow down every request so only do it in development environment
-    if Rails.env == 'development'
+    if Rails.env.development?
```
cjcolvar, line 197: :+1:

cjcolvar (commented):
I think this good so far but I'm still seeing failing tests.

app/controllers/hyrax/articles_controller.rb
```diff
@@ -7,6 +7,8 @@ class ArticlesController < ApplicationController
     # Adds Hyrax behaviors to the controller.
     include Hyrax::WorksControllerBehavior
     include Hyrax::BreadcrumbsForWorks
+    include HykuAddons::WorksControllerAdditionalMimeTypesBehavior
```
okeen, line 10: Do you guys think we could move this line to `Hyrax::WorksControllerBehavior` to avoid repeating it into every works controller?

app/models/concerns/hyku_addons/solr_document_behavior.rb
```diff
@@ -94,5 +94,12 @@ def attributes
         subject: 'keyword_tesim'
       )
     end
+
+    # Work out the reader class from the solr document model
+    def meta_reader_class
+      "Bolognese::Readers::#{@model.instance_variable_get(:@model)}Reader".constantize
+    rescue NameError
+      "Bolognese::Readers::GenericWorkReader".constantize
```
okeen, line 102: Can we simply return a `Bolognese::Readers::GenericWorkReader` here?

app/services/bolognese/writers/ris_writer_behavior.rb
```diff
@@ -0,0 +1,88 @@
...
+            JO: meta.dig("journal_title"),
+            LA: meta.dig("language"),
+            N1: meta.dig("add_info"),
+            UR: meta.dig("official_link"),
+            IS: container.to_h["issue"],
+            VL: container.to_h["volume"],
+            ER: ""
+          }
+            .compact
+            .map { |k, v| v.is_a?(Array) ? v.map { |vi| "#{k}  - #{vi}" }.join("\r\n") : "#{k}  - #{v}" }
```
okeen, line 69: What do you guys think on moving this compact and custom mapping into a small method with a proper name? It would explains what it does and avoid spending 2 mins figuring it out what it does 6 months from now.

app/views/hyrax/base/_citations.html.erb
```diff
@@ -1,7 +1,8 @@
 <div class="citations">
-    <% if Hyrax.config.citations? %>
-      <%= link_to t(".citations"), hyrax.citations_work_path(presenter), id: 'citations', class: 'btn btn-default' %>
-    <% end %>
+	<% if Hyrax.config.citations? %>
+		<%= link_to t(".citations"), hyrax.citations_work_path(presenter), id: 'citations', class: 'btn btn-default' %>
+	<% end %>
 
-    Citations: <%= link_to "Download as RIS (Zotero, Endnote, Mendeley)", main_app.polymorphic_path(presenter.solr_document, format: :ris) %>
+	<%= t("work_type.show.citations.label") %>
```
okeen, line 6: What if we put this into some context tag (span?) to allow referencing it with CSS selectors for styling or JS purposes?

okeen (approved):
I made some small suggestions but it is looking good :D

app/models/concerns/hyku_addons/solr_document_behavior.rb
prdanelli: Unfortunately not, we need to be able to specify unique fields within the readers that match their Work Form.

cjcolvar: I get that for line 100 but couldn't line 102 be simplified?
```suggestion
      Bolognese::Readers::GenericWorkReader
```

app/views/hyrax/base/_citations.html.erb
cjcolvar: That sounds like a good idea, but maybe we could skip that for now and do it when we have a strong need?

app/models/concerns/hyku_addons/solr_document_behavior.rb
prdanelli: Yea, in hindsight, that seems like the obvious solution.

cjcolvar:
@prdanelli Looks like the build is failing right now on rubocop issues.

app/controllers/hyrax/articles_controller.rb
prdanelli: I think I avoided this initially because of the autoloading. But yes, this is a better solution.

prdanelli: However, I'm not sure it works adding it to the engine like: 

`Hyrax::WorksControllerBehavior.include HykuAddons::WorksControllerAdditionalMimeTypesBehavior`

I'm now getting Unknown Format errors.

cjcolvar: Oh, I forgot that `Hyrax::WorksControllerBehavior` isn't something that we're supplying.  How about we make a `HykuAddons::WorksControllerBehavior` where we can gather all of our includes and include that in each controller?  That way it is explicit and we don't have to inject it in an initializer but avoids having to add another include when the next mixin comes along.

app/services/bolognese/readers/time_based_media_article_reader.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+module Bolognese
+  module Readers
+    class TimeBaseMediaArticleReader < GenericWorkReader
```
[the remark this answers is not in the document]
prdanelli: No worries. 

I had been trying to avoid having all of the functionality in the base class, as it becomes very large and pretty hard to follow.

app/controllers/hyrax/articles_controller.rb
prdanelli: @cjcolvar Are you referring to the [Hyrax::My::WorksController](https://github.com/samvera/hyrax/blob/master/app/controllers/hyrax/my/works_controller.rb)? 

I'm not sure where to include this behaviour override.

cjcolvar (approved):
Looks good! :tada:

cjcolvar:
The failing test is unrelated so I'm going to go ahead and merge this and add it to the list of randomly failing tests.

PR #54, opened 2021-01-15
Test DOI minting

cjcolvar:
Use shared specs from `hyrax-doi` and explicitly include hyrax-doi mixin modules instead of running the install generator each time.


cjcolvar:
I'm going to go ahead and merge this so I can get it into a release today.

PR #56, opened 2021-01-19
#AH-106 Adds consistency on deposit forms hints and placeholders.

okeen:


cjcolvar (dismissed):
Looks good.

cjcolvar:
I think the failing test is due to the form changing with the `records/edit_fields` so that the values aren't actually getting persisted.  I didn't explore enough to see why this was but maybe it is an id/name issue on the fields?

cjcolvar (approved):
Looks amazing! :tada:

PR #57, opened 2021-01-20
Fixing unexecuted application JS

prdanelli:


prdanelli:
@cjcolvar I think this should resolve the JS issues in the admin. Ive also removed the place rails ujs was being included for a second time.

cjcolvar (approved):
Awesome! :tada:  Hope this does it! :crossed_fingers:

PR #59, opened 2021-01-20
Adds Pacific Worktypes

BertZZ:


cjcolvar (commented):
This looks really good and I really only have a couple suggestions about code organization and tightening things up.

The approach we started to take with the BL work types is to only include the properties that are common to all work types and then define the rest in the specific work type model classes.  This would tighten up the model making it more apparent which properties are expected and which aren't.  So I'd like to see the `PacificSharedMetadata` module be reduced to only properties shared by all Pacific work types or the module removed entirely and the properties moved onto the specific work type models. Then I think it would be good to remove delegated methods from the presenters which aren't defined on the model.  Do you think this seems reasonable?

BertZZ:
Yeah for sure, I was mostly worries about bloating the models by including fields that didn't need to be there, but yeah I will do some work and tighten it up like you suggest

cjcolvar (approved):
Looks good!  It looks like we should be able to DRY up the two feature helpers but that can come later.

PR #60, opened 2021-01-21
#AH-107 ensures correct orderin of all the fields of the deposit form

okeen:


cjcolvar (approved):
Looks good to me.  I like how this will ensure a standard order across all work types.

okeen:
Yes, that's what I tried to achieve too.

glad you liked it :D

PR #65, opened 2021-01-26
Missing work types

prdanelli:


app/forms/concerns/hyku_addons/work_form.rb
```diff
@@ -17,15 +17,36 @@ def editor_list
     end
 
     class_methods do
```
okeen, line 19: This looks like a really nice clean up. The only downside is it is going to whitelist every attribute for all the work types, right?

okeen (approved):
Looks great :)

app/forms/concerns/hyku_addons/work_form.rb
prdanelli: Yea, but i'm not sure thats an issue personally. It's better than missing some and then losing data IMO.

prdanelli:
@cjcolvar Feel free to merge this monstrosity of a PR. The form fields and model properties are all correct. Presenters will need checking as i've been adding the form fields, which might not always be required.

PR #67, opened 2021-01-26
Bulk import and export through Bulkrax

cjcolvar:
The workflow for importing is:
1. Export from a UP Hyku 1 instance
1. Run the script to transform the CSV
1. Import with the transformed CSV

There are a number of overrides here which hopefully can get resolved in a better way in coordination with changes upstream.  I was finding it really hard to get information when the import fails which is something that we should try to help with upstream.  Sorry no tests on the CSV transformation script.

Here are some random notes I made during the process:
> - date_modified and depositor autoset by Bulkrax/AF - is this okay?
> - admin set is set on the importer (not per row)
> - importer doesn't have any errors when valid_import? returns false (missing required fields)
> - Pacific creator person has a middle name?!?
> - ResourceTypesService needs to subclass from Hyrax::ResourceTypesService?!? (See https://github.com/samvera/hyrax/blob/d39c0d70a0690fce24d4cfe8280945b9034a7c21/app/models/concerns/hyrax/solr_document_behavior.rb#L26 and https://github.com/samvera-labs/bulkrax/blob/170c32916f0e09d389996d0f594390874da769bd/app/matchers/bulkrax/application_matcher.rb#L95)
> - matcher from/to need to be the same?!?
> - Need to review visibility and rights statement; file upload for import didn't actually upload; collection_id needs to become collection; work_type becomes model; need to strip work_type from resource_type
> - Missing Pacific controlled vocabularies is blocking show page of batch imported works
> - Still need to come up with good field mapping (for date parsing specifically?)


app/matchers/hyku_addons/csv_matcher.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+module HykuAddons
+  class CsvMatcher < Bulkrax::CsvMatcher
+    def parse_date_published(src)
+      date_parts = src.split('-')
```
prdanelli, line 5: Might it be better to use `date = Date.edtf(str)` and then `date.year` etc?

app/models/hyku_addons/csv_entry.rb
```diff
@@ -0,0 +1,63 @@
...
+  class CsvEntry < Bulkrax::CsvEntry
+    def self.matcher_class
+      HykuAddons::CsvMatcher
+    end
+
+    # Override to add date and json field handling
+    def build_metadata
+      raise StandardError, 'Record not found' if record.nil?
+
+      raise StandardError, "Missing required elements, required elements are: #{importerexporter.parser.required_elements.join(', ')}" unless importerexporter.parser.required_elements?(record.keys)
```
prdanelli, line 12: Because this line is so long, could you wrap the exception in a condition block rather than having it inline?

prdanelli (approved):

spec/factories/bulkrax_importers.rb
```diff
@@ -0,0 +1,109 @@
...
+    limit { 10 }
+    parser_fields { { 'import_file_path' => 'spec/fixtures/csv/good.csv' } }
+    field_mapping { {} }
+    after :create, &:current_run
+  end
+
+  factory :bulkrax_importer_csv_complex, class: 'Bulkrax::Importer' do
+    name { 'CSV Import' }
+    admin_set_id { 'MyString' }
+    user { FactoryBot.build(:base_user) }
```
prdanelli, line 46: Can this not be `user { build(:base_user) }`?

spec/features/bulkrax_import_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+           #  admin_set_id: admin_set.id,
+           field_mapping: field_mapping,
+           parser_klass: "HykuAddons::CsvParser",
+           parser_fields: { 'import_file_path' => import_batch_file },
+           limit: 0)
+  end
+  let(:import_batch_file) { 'spec/fixtures/csv/pacific_articles.csv' }
+  let(:field_mapping) do
+    {
+      # "title" => {"from"=>["title"], "split"=>true, "parsed"=>false, "if"=>nil, "excluded"=>false},
```
prdanelli, line 21: Outstand commented code left here.

prdanelli (approved):

app/matchers/hyku_addons/csv_matcher.rb
cjcolvar: Possibly.  The dates coming are in a specific format instead of the full possibilities of EDTF dates so I _think_  this is safe for now.

app/models/hyku_addons/csv_entry.rb
cjcolvar: I believe this is actually this way upstream and rubocop doesn't complain so :man_shrugging:.

spec/factories/bulkrax_importers.rb
cjcolvar: True I'll clean that up.

spec/features/bulkrax_import_spec.rb
cjcolvar: I had left this in for a reason before but I can't recall what reason that is now.  I'll either add a comment or remove it in the next PR.

spec/factories/bulkrax_importers.rb
cjcolvar: I copied that in from upstream so I'm inclined to just leave it like it is.  I'll add a note at the top about the origin of these factories.

PR #68, opened 2021-01-26
Allow resource types to be changed per work types

cjcolvar:
Still need to write tests for HykuAddons::QaSelectService, but it is working.


cjcolvar:
Ugh, I don't know why these tests are failing now.  I don't think they're related to this PR.

PR #72, opened 2021-02-01
Adds remaining Pacific Work types

BertZZ:


cjcolvar (commented):
This looks like a nice commit that touches all the right places!

cjcolvar (approved):
I didn't go through this super closely, but it looks good to me.

PR #77, opened 2021-02-03
Add Custom RIS Data Fields

prdanelli:
Some custom BL fields were missing from the RIS reader fields and this PR attempts to resolve that.


app/services/bolognese/readers/base_work_reader.rb
```diff
@@ -58,18 +62,35 @@ def read_work(string: nil, **options)
         @reader_attributes.merge(read_options)
       end
 
+      # Override the parent method as we have other fields
+      def publication_year
```
cjcolvar, line 66: I'm not really excited about overriding this method because it doesn't seem to be following the patterns of bolognese where it's trying to map into the internal data structure.  I'm concerned that this may break other things or that it makes our use of bolognese pointless (which could be okay).  Maybe we could chat about this tomorrow?

prdanelli: Hi @cjcolvar, 

I'm more than happy to change this. I've got an idea how I could achieve this, so I'll give that a go and get you to check it out when you're online :)

prdanelli: @cjcolvar I've refactored it to not override the method, but set the meta value the method relies on.

app/models/concerns/hyku_addons/solr_document_behavior.rb
```diff
@@ -21,7 +21,7 @@ def attributes
     included do
       class_attribute :registered_attributes
       # Pre-seed registered attributes because they have already been created before this module is included
-      self.registered_attributes = [:identifier, :based_near, :based_near_label, :related_url, :resource_type,
+      self.registered_attributes = [:doi, :identifier, :based_near, :based_near_label, :related_url, :resource_type,
```
prdanelli, line 24: @cjcolvar, there might be another way of doing this, but i think there is missing `registered_attribute` - `:doi` in the `SolrDocument::DOIBehavior` concerns. When I added it here the Bolognese `doi` method and data value was populated as I would expect.

cjcolvar: Huh, I'll take a look at this.

cjcolvar: Good catch.  I think this is the right place to add it for now.

cjcolvar:
Sorry about these unrelated failing tests.  :disappointed:  Eventually they'll get fixed.

cjcolvar (approved):
Looks good.

PR #79, opened 2021-02-04
Remove Source from forms

prdanelli:


app/models/report.rb
```diff
@@ -5,6 +5,7 @@ class Report < ActiveFedora::Base
   include Hyrax::DOI::DOIBehavior
   include Hyrax::DOI::DataCiteDOIBehavior
   include ::HykuAddons::WorkBase
+  include Hyrax::HasRendering
```
cjcolvar, line 8: Should this be included in HykuAddons::WorkBase instead of in every model?  Maybe the hyrax-doi includes should be moved there as well?

prdanelli: My bad, thought I removed it. It's not necessary, That must have been left over from when I was trying to work out what was going on.

cjcolvar (commented):
Looks good!  So does this take care of AH-151 and the bug blocking creating ThesisOrDissertation works?  Does it close anything else?

cjcolvar (approved):

PR #80, opened 2021-02-04
Catch error causing 500 on root admin user registration

prdanelli:
On this page, http://lvh.me:3000/?locale=en, i'm getting 500 errors because the `current_account` is `nil`.


cjcolvar (approved):
Good catch! We should probably write a feature test for this eventually.

PR #81, opened 2021-02-05
Fixes to API work endpoint

BertZZ:


spec/requests/api_v1_work_spec.rb
```diff
@@ -112,6 +112,14 @@
...
                                          "workflow_status" => nil)
       end
+
+      it 'returns work json with data when it exists' do
+        abstract = "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi at tincidunt nisl. Nulla et lacus consequat, interdum eros nec, pulvinar."
+        work.abstract = abstract
+        work.save
+        get "/api/v1/tenant/#{account.tenant}/work/#{work.id}"
+        expect(json_response).to include("abstract" => [abstract])
+      end
```
cjcolvar, line 122: I think this might be cleaner in a context:
```suggestion
      context 'with data when it exists' do
        let(:work) { create(:work, visibility: 'open', abstract: abstract) }
        let(:abstract) { 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Morbi at tincidunt nisl. Nulla et lacus consequat, interdum eros nec, pulvinar.' }

        it 'returns work json' do
          get "/api/v1/tenant/#{account.tenant}/work/#{work.id}"
          expect(json_response).to include("abstract" => [abstract])
        end
      end
```

cjcolvar (commented):
This looks good.  Let me know when you think it is ready for a final review.

BertZZ:
> This looks good. Let me know when you think it is ready for a final review.

I was going to wait for Thumbnail and/or file sets were added but those seem hard and maybe not the best place to put energy right now SO i think this is good

cjcolvar (commented):
Looks like spec/requests/api_v1_work_spec.rb:33 needs to be updated to match the newly returned values.

cjcolvar (approved):

PR #83, opened 2021-02-08
AH-152: DOI Form filling

prdanelli:


app/controllers/concerns/hyku_addons/doi_controller_behavior.rb
```diff
@@ -0,0 +1,68 @@
...
+
+      protected
+
+      def json_response
+        {
+          data: formatted_work,
+          curation_concern: params[:curation_concern],
+          fields: {
+            json: curation_concern_class.json_fields,
+            date: curation_concern_class.date_fields,
```
cjcolvar, line 36: I think it would be best not to require the JS to know about these.  I think it might tie the JS too closely to the implementation details on the backend.  How about splitting these fields on the server-side so the JSON response is just the fields that match up with the form fields on the page?

app/controllers/concerns/hyku_addons/doi_controller_behavior.rb
```diff
@@ -0,0 +1,68 @@
...
+        raise Hyrax::DOI::NotFoundError, "DOI (#{doi}) could not be found." if meta.blank? || meta.doi.blank?
+
+        meta.hyrax_work
+      end
+
+      def curation_concern_class
+        curation_concern.constantize
+      end
+
+      def curation_concern_reader_class
```
cjcolvar, line 53: Should this be a writer instead of a reader?

lib/hyku_addons/engine.rb
```diff
@@ -230,6 +230,7 @@ def self.dynamically_include_mixins
       User.include HykuAddons::UserEmailFormat
       Bolognese::Writers::RisWriter.include Bolognese::Writers::RisWriterBehavior
       Hyrax::GenericWorksController.include HykuAddons::WorksControllerBehavior
+      Hyrax::DOI::HyraxDOIController.include HykuAddons::DOIControllerBehavior
```
cjcolvar, line 233: If we come up with a new pattern that we like and is generic, I think it would be good to push it back upstream to `hyrax-doi` and hopefully remove this override eventually.

cjcolvar (commented):
The JS looks good at a quick glance.  We can chat tomorrow if any of my comments don't make sense. :slightly_smiling_face:

app/controllers/concerns/hyku_addons/doi_controller_behavior.rb
prdanelli: @cjcolvar would you be able to clarify this one for me?

app/models/concerns/hyku_addons/generic_work_overrides.rb
```diff
@@ -124,7 +124,7 @@ module GenericWorkOverrides
       end
 
       self.json_fields += %i[editor current_he_institution]
-      self.date_fields += %i[event_date related_exhibition_date]
+      self.date_fields += %i[event_date related_exhibition_date date_created]
```
cjcolvar, line 127: I don't believe that `date_created` is a field on our models. (It might be in upstream GenericWork but not on ours.)

cjcolvar (commented):
I think a lot of this work should be pushed into a Bolognese writer module that creates work objects using our work types instead of upstream Hyku's GenericWork.  If it creates a work object that matches the requested `model_class` then the JSON fields should only need to be parsed from the string form and included in the response.

app/models/concerns/hyku_addons/generic_work_overrides.rb
prdanelli: You are correct. I added it as a test because `hyrax_work` is only giving me `date_created` in the json... and I forgot to remove it.

app/controllers/concerns/hyku_addons/doi_controller_behavior.rb
```diff
@@ -0,0 +1,65 @@
...
+            curation_concern: curation_concern
+          }.to_json
+        end
+
+        def formatted_work
+          # Download the requested DOI response
+          input = reader_class.new(input: doi)
+
+          # Use that response to build the object as we would expect it
+          meta = reader_class.new(input: input.string)
```
prdanelli, line 43: @cjcolvar There might be a better way of doing this, if so, could you give me your thoughts?

The DOI request returns a partially formed reader object, with `string` being assigned the result of the response. However the readers all assume that `meta` is available and so the above gets around that by rebuilding the reader with the XML string.

prdanelli:
@cjcolvar this is ready to review now. I came to the conclusion that we don't really need to know which fields are on the form/model. Everything seems to be just fine without passing `creation_concern` as the `work_type`, so i've removed the extra complexity. 

If you think we might need to know this in the future, I would always roll back the commit in which i removed it.

cjcolvar (approved):
:tada: Let's get this deployed! :shipit:

cjcolvar (approved):

PR #84, opened 2021-02-10
Creator and contributor need to be added to the override

BertZZ:
I forgot yesterday :sweat_smile:

### DIFF
--- a/app/views/hyku/api/v1/work/_work.json.jbuilder
+++ b/app/views/hyku/api/v1/work/_work.json.jbuilder
@@ -15,14 +15,10 @@ json.challenged work.try(:challenged)
 json.cname @account.cname
 #                                         "collections" => nil,
 creator = work.creator.try(:first)
-if creator
-  json.creator JSON.parse(creator)
-end
+json.creator creator.present? ? JSON.parse(creator) : []
 
 contributor = work.contributor.try(:first)
-if contributor
- json.contributor JSON.parse(contributor)
-end
+json.contributor contributor.present? ? JSON.parse(creator) : []
 #                                         "current_he_institution" => nil,
 #                                         "date_accepted" => nil,
 json.date_published work.try(:date_published)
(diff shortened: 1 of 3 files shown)


app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -14,6 +14,8 @@ json.buy_book work.try(:buy_book)
 json.challenged work.try(:challenged)
 json.cname @account.cname
 #                                         "collections" => nil,
+json.contributor work.contributor
+json.creator work.creator
```
cjcolvar, line 18: Do these need to be `contributor_display` and `creator_display` instead?  `contributor` and `creator` hold the JSON blob.

BertZZ: Oooop. Ill check

BertZZ: `creator_display` is not a method. We can do `JSON.parse(creator)`

app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -14,6 +14,15 @@ json.buy_book work.try(:buy_book)
 json.challenged work.try(:challenged)
 json.cname @account.cname
 #                                         "collections" => nil,
+creator = work.creator.try(:first)
+if creator
+  json.creator JSON.parse(creator)
+end
```
cjcolvar, line 20: For these JSON fields, do you think we need to return an empty array in the response JSON if no values exist?
```suggestion
json.creator creator.present? ? JSON.parse(creator) : []
```

cjcolvar (commented):
Looks like there are some rubocop violations.

cjcolvar (approved):
This looks good.  I'm going to merge it now, but I wonder if there are any other JSON fields that need to have the same treatment.

PR #87, opened 2021-02-15
Feature/settings

prdanelli:


app/controllers/hyku_addons/account_settings_controller.rb
```diff
@@ -35,15 +35,9 @@ def update_single
...
-          settings: [:contact_email, :index_record_to_shared_search, :live, :enabled_doi, :gtm_id,
-                     :turn_off_fedora_collection_work_association, :file_size_limit, :enable_oai_metadata,
-                     :add_collection_list_form_display, :hide_form_relationship_tab, :shared_login,
-                     :redirect_on, :institutional_relationship_picklist, :institutional_relationship,
-                     :oai_prefix, :oai_sample_identifier, :oai_admin_email, :allow_signup,
-                     email_format: [], help_texts: {}, work_unwanted_fields: {}, work_type_list: [],
-                     required_json_property: {}, contributor_roles: [], metadata_labels: {},
-                     creator_roles: [], html_required: {}, licence_list: [:name, :url],
-                     weekly_email_list: [], monthly_email_list: [], yearly_email_list: []]
+          settings: [:contact_email, :gtm_id, :file_size_limit, :enable_oai_metadata,
```
cjcolvar, line 38: I thought we were going to get rid of `:enable_oai_metadata` because it is going into a feature flipper.

app/views/hyku_addons/account_settings/edit.html.erb
```diff
@@ -1,57 +1,42 @@
...
+  <br>
+  <%= link_to "Back", hyku_addons.admin_account_settings_path, class: "btn btn-xs btn-default" %><br>
+  <h1><span class="fa fa-gears"></span>Editing Account Settings</h1>
+  <br>
+<% end %>
+
+<table class="table table-striped">
+  <tbody>
+    <%= form_for(@account, url: hyku_addons.admin_account_setting_path(@account), html: { class: 'form control' }) do |f| %>
+      <% ['contact_email', 'gtm_id'].each do |fname| %>
```
cjcolvar, line 11: I think extracting these lists of settings (here and on lines 17 and 33 ) into AccountBehavior as something like constants could clean things up and avoid having the views knowing about specific settings when they don't really need to.  What do you think?

cjcolvar (approved):
This looks great!  I had a couple questions but nothing that would block this from being merged now.

app/controllers/hyku_addons/account_settings_controller.rb
prdanelli: @cjcolvar its not on the spreadsheet I was working from. I can remove it though.

PR #88, opened 2021-02-16
Add WorksControllerBehaviour to new controllers

prdanelli:
I removed the extra unnecessary comment lines from the controllers whilst I was adding the missing concern - just boy scouting to make a nice code environment


cjcolvar (approved):
Good catch!

PR #90, opened 2021-02-17
Feedback on Pacific worktypes

BertZZ:


app/forms/concerns/hyku_addons/work_form.rb
```diff
@@ -41,7 +41,7 @@ def build_permitted_params
       # @param work_type_terms [Array] array of terms to add to the work type
       def add_terms(work_type_terms = [])
         self.terms ||= hyrax_terms
-        self.terms = available_terms & (self.terms + Array.wrap(work_type_terms))
+        self.terms = (self.terms + Array.wrap(work_type_terms)) & available_terms
```
cjcolvar, line 44: Can you explain why you're making this change?  It looks like it will change the preference of the ordering of fields on the deposit form to be the form's over `available_terms`.  The reason it was the other way was to standardize the order of the fields on deposit forms across work types.  But maybe the need to allow fields to be in different orders for different work types is greater than the desire for a standardized order.  Thoughts?

cjcolvar (commented):
This is looking good so far!

app/forms/concerns/hyku_addons/work_form.rb
BertZZ: It is exactly for that reason. Especially when not all wortk types will have all fields, having the order defined here would just start to get messy. Seems to me it would be better to just let the forms define the order themselves

cjcolvar: Yeah, I can see it both ways.  We'll have to go through and ensure the ordering is what we want for all existing work types if we change this.
I'd be interested in hearing from @okeen because he wrote this originally.

okeen: If we need to have specific ordering for some work types (let's call them Exceptions), we will need to get rid of the automatic ordering or allow overwriting the ordering per Exception. I guess the condition to use one approach or the other will depend on the number of Exceptions

cjcolvar:
I noticed that there were some different subfields for creator/contributor when I was looking at a pacific export so I'm glad to see them getting taking care of since I haven't written them up yet.  Instead of making a new mixin and special methods to handle these, do you think we could make them standard across all models?  This is a question for Elisa and Tom of course.

cjcolvar (approved):
This looks good to me and I think it should work once the authority files are copied as part of the hyku_addons:install generator.  I was running into issue on my local setup so I wasn't able to verify in the UI but it was working from the rails console so I'm going to merge and we can test on demo.

BertZZ:
Ok keep an eye out for the error I sent you if you couldn't reproduce locally. It will occur when you try and view the PacificArticle form

PR #92, opened 2021-02-18
Adds a rake task to update the frontend_url of accounts

okeen:


lib/tasks/account.rake
```diff
@@ -6,5 +6,10 @@ namespace :hyku do
     task :create, [:name] => [:environment] do |_t, args|
       CreateAccount.new(Account.new(name: args[:name])).save
     end
+    desc 'Update the frontend url of an account'
+    task :frontend_url, [:tenant, :frontend_url] => [:environment] do |_t, args|
+      account = Account.find_by(tenant: args[:tenant])
+      UpdateAccountFrontendUrl.new(account, args[:frontend_url]).save
```
cjcolvar, line 12: I think it would be useful to have some feedback if this operation succeeded or not.

app/services/hyku_addons/update_account_frontend_url.rb
```diff
@@ -0,0 +1,40 @@
...
+  # @param account [Account]
+  # @param frontend_url [String] Must be a valid URI
+  def initialize(account, frontend_url)
+    @account = account
+    @frontend_url = frontend_url
+  end
+
+  ##
+  # @return [Boolean] true if validations pass and was able to update the account
+  def save
```
cjcolvar, line 19: What do you think about renaming this to `run` or `perform` or something like that?  I feel like `save` should be reserved for persistable objects like ActiveRecord models.  I'm open to anything though and don't want to block this over the method naming.

cjcolvar (dismissed):
Looking good just had a couple questions/suggestions that shouldn't block this being merged.

lib/tasks/account.rake
okeen: That's a good point. I will make that happen

app/services/hyku_addons/update_account_frontend_url.rb
okeen: I think it's a good idea. I just followed an example Service that I found inside the internal test app and used 'save' to keep a bit of consistency and it is, in the end, saving some changes on the Account model, but it feels like 'perform' is a much better name for what Service objects do

cjcolvar: Interesting.  I didn't know any service classes in Hyku were like that.  The examples I've seen are more like https://github.com/samvera/hydra-works/blob/master/lib/hydra/works/services/upload_file_to_file_set.rb

cjcolvar (approved):

PR #97, opened 2021-02-18
AH-174: I18n Multitenant configuration

prdanelli:
This Feature branch enables us to use the repo name in the locale yml file name in order to override the specific translations for a specific client, thereby customising areas of the admin without changing the templates.

If the repo had the name of "Test", then: 

For the "en" locale, the file would be called "en-TEST.yml"
For Brazilian Portuguese, the file would be called "pt-BR-TEST.yml"


lib/hyku_addons/engine.rb
```diff
@@ -10,7 +10,7 @@ class Engine < ::Rails::Engine
       # Eager load required for overrides in the initializer below
       # There is probably a better solution for this but I don't think it is worth the time
       # tinkering with it since this works and doesn't cause too much of a slowdown
-      if Rails.env == 'development' # Only do this for development environment for now
+      if Rails.env.development?
```
cjcolvar, line 13: :+1:

okeen (approved):
looking great

PR #101, opened 2021-02-22
Handle creator and contributor multi-part fields for collections

cjcolvar:
Collection has creator and contributor fields which were matching the edit partials (e.g. `records/edit_fields/_creator.html.erb`) causing a NoMethodError on `creator_list`.  Looking at Hyku 1, these fields should be multi-part JSON fields like for works so I made them work that way.

Fixes AH-169.


prdanelli (approved):

okeen (approved):
Looks great!

PR #103, opened 2021-02-23
AH-177: Institutional Relationship QA Service and Pacific Tenant Config

prdanelli:


app/views/records/edit_fields/_creator.html.erb
```diff
@@ -43,7 +44,7 @@
 				{ field_type: :text, field_slug: :creator_family_name, field_args: { data: { required: true, required_group: :creator_name, on_blur: :toggle_required_group } } },
 				{ field_type: :text, field_slug: :creator_given_name, field_args: { data: { required: true, required_group: :creator_name, on_blur: :toggle_required_group } } },
 				{ field_type: :text, field_slug: :creator_orcid },
-				{ field_type: :select, field_slug: :creator_institutional_relationship, select_options: ['Staff member', 'Research associate'], field_args: { multiple: true } }] %>
+				{ field_type: :select, field_slug: :creator_institutional_relationship, select_options: HykuAddons::InstitutionalRelationshipService.new.select_active_options, field_args: { multiple: true } }] %>
```
cjcolvar, line 47: Do you want to pass the model along to the QA service?

prdanelli: @cjcolvar thats done.

config/locales/en-PACIFIC.yml
```diff
@@ -0,0 +1,8 @@
+en:
```
cjcolvar, line 1: Is this file meant to be added?  I would expect it to come after #97

prdanelli: @cjcolvar It's in preparation for #97. I know I need the files, but there might be more work required on the other PR in the short term, so it's setup for when its ready.

cjcolvar: @prdanelli I added a commit to #97 and did some testing.  Looks like this line should be changed otherwise it will override the default `:en` locale.
```suggestion
en-PACIFIC:
```

prdanelli: Thanks @cjcolvar, I've not worked this deep inside the I18n before, so this is useful to know.

okeen (approved):
Looks OK mate 👍

PR #107, opened 2021-02-24
AH-170  Improve tenant automation with rake tasks

okeen:
Adds rake tasks to delete tenants and allows tenant creation to pass array of admin emails

### DIFF
--- a/lib/tasks/account.rake
+++ b/lib/tasks/account.rake
@@ -3,8 +3,15 @@
 namespace :hyku do
   namespace :account do
     desc 'Create an account'
-    task :create, [:name] => [:environment] do |_t, args|
-      CreateAccount.new(Account.new(name: args[:name])).save
+    task :create, [:name, :admin_emails] => [:environment] do |_t, args|
+      account = Account.new(name: args[:name])
+      CreateAccount.new(account).save
+      AccountElevator.switch!(account.cname)
+      Array.wrap(args[:admin_emails]).each do |admin_email|
+        User.invite!(email: admin_email) do |u|
+          u.add_role(:admin)
+        end
+      end
     end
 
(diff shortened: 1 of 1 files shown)


prdanelli (approved):

lib/tasks/account.rake
```diff
@@ -16,5 +23,33 @@ namespace :hyku do
...
+end
+
+def check_confirmation
+  unless ENV['CONFIRM'] == 'yes'
+    $stderr.puts <<-EOC
+WARNING: This process will destroy all data for this tenant in:
+DB: All tables
+Fedora: #{ActiveFedora.fedora.build_connection.http.url_prefix}
+Solr: #{ActiveFedora.solr_config[:url]}
+Redis: #{Sidekiq.redis { |c| c._client.options.values_at(:host, :port, :db).join(':') }}
```
cjcolvar, line 49: Are these accurate for the tenant or do you need to `AccountElevator.switch!` before doing this?

okeen: good point, it only shows the generic connection info right now. I will make another PR for it.

lib/tasks/account.rake
```diff
@@ -16,5 +23,33 @@ namespace :hyku do
...
+Please run `rake app:hyku::account::cleanup[{name}] CONFIRM=yes` to confirm.
```
cjcolvar, line 50: It is super confusing but this should be `hyku_account::cleanup`.  Within `hyku_addons` the rake task is prepended with `app` because of the two different contexts that rake tasks can be run: the engine itself and the internal test hyku application.  Outside of development on `hyku_addons` and in the context of a running hyku application the `app` namespace won't exist.

PR #108, opened 2021-02-24
hides to contributor role field for pacific work types

BertZZ:
Hides contributor role for Pacific worktypes as per line 31 of [https://docs.google.com/spreadsheets/d/1wyBsV3E-Onwgvi1VPWRrRUJ3g6sk5QvhhFe1dKHM92k](url)


cjcolvar (approved):
This conditional feels a bit icky to me.  Ideally I think I'd like to see this in a partial that takes precedence over this one for the specific work types, but I'm not sure if that is possible with our use of `records/edit_fields`.  Maybe it is about time that we extract the pacific work types into their own plugin?

PR #110, opened 2021-02-24
Improves user feedback on account rake tasks

okeen:


lib/tasks/account.rake
```diff
@@ -15,9 +15,10 @@ namespace :hyku do
     end
 
     desc 'destroy an account and all the data within'
-    task :cleanup, [:name] => [:environment] do |_t, args|
-      account = Account.find_by(name: args[:name])
-      if account.present? && check_confirmation
+    task :cleanup, [:tenant] => [:environment] do |_t, args|
+      account = load_account(args[:tenant])
+      AccountElevator.switch!(account.cname)
```
cjcolvar, line 20: Is this switch even necessary?  I think it could maybe be removed.

okeen: true, it is not necessary

cjcolvar (approved):
Looks good!

PR #111, opened 2021-02-24
AH-185: Toggle DOI

prdanelli:
Add tenant wide settings for toggling DOI minting.


cjcolvar:
I think this would better live in `hyrax-doi`.  Could you add it to your PR there?

prdanelli:
This is a testing branch used before changes were moved to hyrax-doi.

PR #116, opened 2021-02-25
Pacific deposit form changes

BertZZ:
Overides Title and Volume to make them singular fields and adds further customisations to deposit form labels


cjcolvar (commented):
Is it possible to write these using form helpers?  Something like:
`<%= f.input :title, as: :text, required: f.required?(:title) %>`

cjcolvar (approved):

PR #118, opened 2021-02-25
Admin collection required fields

prdanelli:
Fix an issue with the JS where by all data-required fields were marked as required, even if the attribute was set to `false`.

### DIFF
--- a/app/views/records/edit_fields/_creator.html.erb
+++ b/app/views/records/edit_fields/_creator.html.erb
@@ -29,7 +29,7 @@
 		<div data-toggleable-group="Organisational">
 			<% # TODO: workout why setting the name fields as required causes the deposit form to not register the meta data was entered %>
 			<% [
-				{ f: f, field_type: :text, field_slug: :creator_organization_name, field_args: { data: { required: f.required?(:creator) } } },
+				{ f: f, field_type: :text, field_slug: :creator_organization_name, field_args: { data: { required: f.object.required?(:creator) } } },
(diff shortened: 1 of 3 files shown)


app/views/records/edit_fields/_creator.html.erb
```diff
@@ -1,5 +1,6 @@
 <% template = f.object.model.class.to_s.underscore %>
 <% types = HykuAddons::NameTypeService.new.select_active_options.flatten.uniq! %>
+<% required = f.object.class != Hyrax::Forms::CollectionForm %>
```
cjcolvar, line 3: I know I suggested the check against the class before but I think it would be better to ask the form:
```suggestion
<% required = f.required?(:creator) %>
```

cjcolvar (commented):
This looks good.  There was one thing I suggested that I think would be a good change before merging.

app/views/records/edit_fields/_creator.html.erb
prdanelli: The change is made. I opted to use the `f.required?(:creator)` inline to make it a little more self explaining.

cjcolvar:
Sorry I think I gave the wrong suggestion.  Try `f.object.required?(:creator)` instead.

prdanelli:
> Sorry I think I gave the wrong suggestion. Try `f.object.required?(:creator)` instead.

Nah, its my fault for not testing it after I did it... just assumed it was correct :)

cjcolvar (approved):

PR #119, opened 2021-02-26
AH-179: API Settings

prdanelli:
Add settings to tenant api endpoint and include Google Scholarly Work Types

### DIFF
--- a/app/views/hyku_addons/account_settings/settings/_render_select.html.erb
+++ b/app/views/hyku_addons/account_settings/settings/_render_select.html.erb
@@ -1,3 +1,8 @@
 <%= render 'hyku_addons/account_settings/settings/partial_for_select',
-field_name: @field_name, html_name_element: @field_name, account: @account, display_save_button: true  %>
+    options: options,
+    field_name: @field_name,
+    html_name_element: @field_name,
+    account: @account,
+    display_save_button: true
+ %>
 
--- a/app/views/hyku_addons/account_settings/settings/_render_select.html.erb
+++ b/app/views/hyku_addons/account_settings/settings/_render_select.html.erb
@@ -1,0 +1,3 @@
+<%= render 'hyku_addons/account_settings/settings/partial_for_select',
+field_name: @field_name, html_name_element: @field_name, account: @account, display_save_button: true  %>
+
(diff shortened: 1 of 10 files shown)


app/views/hyku_addons/account_settings/settings/_render_select.html.erb
```diff
@@ -0,0 +1,8 @@
+<%= render 'hyku_addons/account_settings/settings/partial_for_select',
```
cjcolvar, line 1: Should this be `partial_for_select_fields`?  I don't see a `partial_for_select` anywhere.

prdanelli: Well spotted. This file is actually not required at all and must have been left over from my first attempt and working out what was going on.

cjcolvar (commented):
This looks good.  Do you think we should add a test that the settings are being returned in the tenant api endpoint?  Seems like it might be a good idea.

cjcolvar:
I think the failing test is a flaky one that doesn't need to be bothered with as part of this PR.

prdanelli:
> I think the failing test is a flaky one that doesn't need to be bothered with as part of this PR.

No worries, this is done.

spec/requests/v1/tenant_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+    Apartment::Tenant.switch(account.tenant) { Site.update(account: account) }
+  end
+
+  after do
+    WebMock.enable!
+    Apartment::Tenant.drop(account.tenant)
+  end
+
+  describe "/tenant/:id" do
+    let(:work_types) { ["Article", "Book", "ThesisOrDissertation", "BookChapter"] }
```
cjcolvar, line 21: I think I'm missing something.  Where does this list get set on the account?

prdanelli: Hi Chris. They were defined in the Account Factory. I've moved them into the spec to make them more self documenting though and added an extra spec to prove that they are outputting the result of the account settings.

cjcolvar (commented):
This looks good to me, but there are a few rubocop issues that are blocking this.  I'm happy to merge this once those are cleared up and the tests pass.

prdanelli:
Those are all fixed now.

cjcolvar:
@prdanelli I'm still seeing a couple rubocop issues in the CircleCI tests: https://app.circleci.com/pipelines/github/ubiquitypress/hyku_addons/742/workflows/4a90eb69-f19f-4aeb-bcd5-3a0736942725/jobs/1062/parallel-runs/0/steps/0-110?invite=true

prdanelli:
@cjcolvar sorry, must have been not paying attention. Defo fixed now.

cjcolvar (approved):
:tada:

PR #121, opened 2021-03-01
Adds cname to account creation task

okeen:


cjcolvar (approved):
Not blocking this at all, but I wonder if moving forward a different approach for passing params would be good as we're having more than 2 and some optional.  Maybe optsparser or something similar could solve that?

PR #122, opened 2021-03-01
AH-172: Toggle Relationship Tab for admin set

prdanelli:


config/features.rb
```diff
@@ -3,4 +3,8 @@
   feature :oai_endpoint,
           default: true,
           description: "Enable OAI-PMH endpoint for harvesting."
+
+  feature :revised_admin_set_layout,
+          default: false,
+          description: "Move the admin set to the works modal and hide the relationships tab on the deposit form"
```
cjcolvar, line 9: Naming is probably going to be hard for this PR.  Thanks for your first pass and here's mine.  What do you think about this for the name of the feature and description?
```suggestion
  feature :simplified_admin_set_selection,
          default: false,
          description: "The admin set to deposit works is selected from the add work modal and the relationships tab on the deposit form is hidden for depositors."
```

app/helpers/hyku_addons/revised_admin_set_work_form_helper.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  module RevisedAdminSetWorkFormHelper
```
cjcolvar, line 4: How about renaming this?
```suggestion
  module SimplifiedAdminSetSelectionWorkFormHelper
```

cjcolvar (commented):
I added some naming suggestions.  It might be good for Elisa or someone else to take a look and give feedback.

config/features.rb
prdanelli: @cjcolvar I had no idea what to call it, so i'm glad of the improvement :)

spec/helpers/simplified_admin_set_selection_work_form_helper_spec.rb
```diff
@@ -16,7 +16,7 @@
   end
   let(:form) { form_class.new(work, nil, nil) }
   let(:helper) do
-    _view.tab do |v|
+    _view.tap do |v|
```
cjcolvar, line 19: :laughing: I totally missed this too!

spec/helpers/simplified_admin_set_selection_work_form_helper_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      include ::HykuAddons::GenericWorkOverrides
+    end
+  end
+  let(:work) { model_class.new(title: ["Moomin"], depositor: user.user_key) }
+  let(:form_class) do
+    Class.new(Hyrax::GenericWorkForm) do
+      include ::HykuAddons::GenericWorkFormOverrides
+    end
+  end
+  let(:form) { form_class.new(work, nil, nil) }
```
cjcolvar, line 20: I think this could be simplified since it is super complex in `hyrax-doi`.
```suggestion
  let(:work) { GenericWork.new(title: ["Moomin"], depositor: user.user_key) }
  let(:form) { Hyrax::GenericWorkForm.new(work, nil, nil) }
```

spec/helpers/simplified_admin_set_selection_work_form_helper_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      before do
+        allow(Flipflop).to receive(:enabled?).with(:simplified_admin_set_selection).and_return(true)
+        allow(helper).to receive(:can_edit?).with(form.model).and_return(true)
+        allow(helper).to receive(:depositor?).with(form.depositor).and_return(true)
+      end
+
+      it "includes relationships" do
+        expect(helper.form_tabs_for(form: form)).not_to include("relationships")
+      end
+    end
```
cjcolvar, line 44: I think it would be good to add another test here for the other permutations `can_edit? = false` and `can_edit? = true && depositor? = false`.

cjcolvar (commented):
I remember before you saying that the css wasn't coming in so you switched to doing it inline.  Did that get resolved?  I have some ideas if it is still a problem.

prdanelli:
> I remember before you saying that the css wasn't coming in so you switched to doing it inline. Did that get resolved? I have some ideas if it is still a problem.

It is still a problem that I never managed to figure out.

app/forms/concerns/hyku_addons/generic_work_form_overrides.rb
```diff
@@ -23,7 +23,13 @@ module GenericWorkFormOverrides
...
 
       def primary_terms
-        super - %i[license]
+        primary_terms = super - %i[license]
+
+        if Flipflop.enabled?(:simplified_admin_set_selection)
+          primary_terms + %i[admin_set_id]
+        else
+          primary_terms
+        end
```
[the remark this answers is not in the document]
prdanelli: @cjcolvar without this the admin_set_id field isn't added to the form. I think because this is a concern and the WorkForm is a concern.

cjcolvar:
> > I remember before you saying that the css wasn't coming in so you switched to doing it inline. Did that get resolved? I have some ideas if it is still a problem.
> 
> It is still a problem that I never managed to figure out.

:+1: I think if you move the application.css back under hyku_addons then something can be added to the install generator to inject this into the hyku application like for JS (https://github.com/ubiquitypress/hyku_addons/blob/c5929aa081a46b58ab8cc01d244aebaa70b32aa7/lib/generators/hyku_addons/install_generator.rb#L37-L41).  We probably want it injected after hyrax just like the JS.

spec/helpers/simplified_admin_set_selection_work_form_helper_spec.rb
```diff
@@ -32,6 +32,30 @@
...
+
+    context "when the user can not edit the work and is not the depositor" do
+      before do
+        allow(Flipflop).to receive(:enabled?).with(:simplified_admin_set_selection).and_return(true)
+        allow(helper).to receive(:can_edit?).with(form.model).and_return(false)
+        allow(helper).to receive(:depositor?).with(form.depositor).and_return(false)
+      end
+
+      it "includes relationships" do
+        expect(helper.form_tabs_for(form: form)).to include("relationships")
```
cjcolvar, line 44: This should be not include, right?

prdanelli:
@cjcolvar These are all taken care of now.

cjcolvar (approved):
:tada: Looks good!

PR #123, opened 2021-03-01
Bulkrax improvements

cjcolvar:
- Create admin sets (as necessary) and associate works with admin sets (based upon admin set title)
- Transform DOIs into acceptable format in csv preparation script


app/models/hyku_addons/csv_entry.rb
```diff
@@ -31,6 +31,12 @@ def build_metadata
       parsed_metadata
     end
 
+    # Override to use the value in prefer the named admin set provided in the admin_set column then fallback to previous behavior
+    def add_admin_set_id
+      parsed_metadata['admin_set_id'] = AdminSet.where(title: raw_metadata['admin_set']).first&.id if parsed_metadata['admin_set_id'].blank?
+      parsed_metadata['admin_set_id'] = importerexporter.admin_set_id if parsed_metadata['admin_set_id'].blank?
```
okeen, line 37: This looks like could be done in a single assignation sentence, WDYT?
```suggestion
      parsed_metadata['admin_set_id'] ||= AdminSet.where(title: raw_metadata['admin_set']).first&.id || importerexporter.admin_set_id 
```

okeen (dismissed):
Looks great! Nothing blocking, just a small code suggestion

app/models/hyku_addons/csv_entry.rb
cjcolvar: I was thinking the same, but I think it is possible for parsed_metadata['asmin_set_id'] to be empty string at this point so `||=` wouldn't catch it.  But maybe the start of this method could just delete it then we could do the `||=`. 
```
parsed_metadata.delete('admin_set_id') if parsed_metadata['admin_set_id'].blank?
parsed_metadata['admin_set_id'] ||= AdminSet.where(title: raw_metadata['admin_set']).first&.id || importerexporter.admin_set_id 
```

cjcolvar:
I'm going to go ahead and merge this and test it out.

PR #124, opened 2021-03-03
AH-176: Fix Contributor Role for pacific work types

prdanelli:
Rename field in pacific worktypes and add new authority service


okeen (approved):
Looking good :)

okeen (approved):

cjcolvar:
@prdanelli This looks good.  I think if https://github.com/ubiquitypress/hyku_addons/blob/main/config/authorities/contributor_group.yml gets renamed too then CI will go green and it will be good for merging.

prdanelli:
> @prdanelli This looks good. I think if https://github.com/ubiquitypress/hyku_addons/blob/main/config/authorities/contributor_group.yml gets renamed too then CI will go green and it will be good for merging.

Hi @cjcolvar, I dont thnk the original `contributor_group.yml` doesn't need to be changed as this relates to a field that was renamed... it's all a little confusing, but I think it can stay and be used as a default for the normal HykuAddons fields.

cjcolvar:
@prdanelli Where is `contributor_group` used now?  Or is that something entirely different?

cjcolvar (approved):

PR #131, opened 2021-03-05
AH-222: Match Collection to AdminSet

prdanelli:


cjcolvar (commented):
I'm concerned about tying this Pacific requirement too closely to the `simplified_admin_set_selection` feature because I don't think it will always be the case that there is a collection matching the title of the admin set and a tenant might want the simplified selection without the automatic inclusion in collections.  So I think I'd rather see it as a new flipflop feature.  (Maybe `link_admin_sets_and_collections`?)  Does that make sense to you or am I over complicating things?  Maybe my suggestion is flawed since it would be a subfeature given that I think it would only work if `simplified_admin_set_selection` was already enabled.

prdanelli:
> I'm concerned about tying this Pacific requirement too closely to the `simplified_admin_set_selection` feature because I don't think it will always be the case that there is a collection matching the title of the admin set and a tenant might want the simplified selection without the automatic inclusion in collections. So I think I'd rather see it as a new flipflop feature. (Maybe `link_admin_sets_and_collections`?) Does that make sense to you or am I over complicating things? Maybe my suggestion is flawed since it would be a subfeature given that I think it would only work if `simplified_admin_set_selection` was already enabled.

I think because the secondary flipflop would rely on the first it might be tricky to pull this off. 

For example, if the user had "simplified_admin_set_selection" to `true` and "link_admin_sets_and_collections" to `false` we would need to add back the relationships tab and prevent the collections selection inputs. 

If they were both false it would be default Hyku
If 'link_admin_sets_and_collections' is true and 'simplified_admin_set_selection' we need a new interface or js function to link them. 

I understand why you think this could cause issues though. However with simple on/off toggles it is hard to set secondary options or preferences.

cjcolvar (approved):
This looks good...let's test it out and we can always come back and make changes later if necessary.
Hopefully this JS change fixes our JS woes! :crossed_fingers:

PR #134, opened 2021-03-05
Creator switcheroo

cjcolvar:
Switch SolrDocument attributes so `creator` returns the human-readable form instead of the JSON.  Instead expose the JSON as `creator_json`.  This should help us avoid having to override Hyrax views which expect a displayable string for `creator`.


prdanelli:
We're keeping this around for now as it might be a good idea to pick up again later

prdanelli:
I would love this to be finished but it would need a substantial migration of client data now.

PR #140, opened 2021-03-09
Update installation instructions

prdanelli:


cjcolvar (approved):
Thanks for writing these up!

PR #141, opened 2021-03-09
Try to fix flakey feature test failures by being patient

cjcolvar:
This is what was adopting in upstream hyku to deal with this issue.
See https://github.com/samvera/hyku/pull/1668


cjcolvar:
That didn't seem to help. :cry:

PR #147, opened 2021-03-10
Removes Pacific from labels

BertZZ:


BertZZ:
The new work forms don't use the locale files so will still say `Create New Pacific xxxxxx` until this is fixed

cjcolvar (approved):
Looks good!

PR #148, opened 2021-03-10
Bulkrax Improvements

cjcolvar:
Fix issues with admin sets, collections, parsing of multiple valued fields, date fields, and resource types.


app/models/hyku_addons/csv_entry.rb
```diff
@@ -59,11 +78,31 @@ def add_json_fields
       end
     end
 
+    def add_resource_type
+      resource_type_service = HykuAddons::ResourceTypesService.new(model: parsed_metadata['model'].safe_constantize)
```
okeen, line 82: Will this need a safe operator? 

`parsed_metadata['model']&.safe_constantize`

app/parsers/hyku_addons/csv_parser.rb
```diff
@@ -1,6 +1,12 @@
 # frozen_string_literal: true
 module HykuAddons
   class CsvParser < Bulkrax::CsvParser
+    # FIXME: Override to make debugging easier
+    def perform_method
+      return :perform_now if Rails.env == 'development' || Rails.env == 'test'
```
okeen, line 6: maybe I would replace that if for:

```suggestion
      return :perform_now unless Rails.env.production?
```

WDYT?

okeen (approved):

app/parsers/hyku_addons/csv_parser.rb
cjcolvar: Yes, I forgot about that.  I'll get that in the next PR.

app/models/hyku_addons/csv_entry.rb
cjcolvar: I think it should always be there but wouldn't hurt to add and might help with debugging.  I'll get it in the next PR.

PR #158, opened 2021-03-12
AH-226: Add file set download route, which returns JSON for frontend

prdanelli:
I can't test this locally as it requires GC configuration that just wont work outside of the production environment. I'd like to merge it in and get it onto production to make sure its all working correctly.

### DIFF
--- a/app/controllers/concerns/hyku_addons/file_sets_controller_behavior.rb
+++ b/app/controllers/concerns/hyku_addons/file_sets_controller_behavior.rb
@@ -1,0 +1,12 @@
+module HykuAddons
+  module FileSetsControllerBehavior
+    extend ActiveSupport::Concern
+
+    def download
+      render json: {
+        signed_url: HykuAddons::FileSignedUrlService.new(file: @file_set.original_file).perform,
+        file_name: @file_set.title.first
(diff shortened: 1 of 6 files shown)


cjcolvar (commented):
I think that I'd like to see this in the API because it is necessary to power the front-end and to keep it isolated from the rest of Hyku/Hyrax because there are different mechanisms there to download files and generate single use links.  I think the code should still live here in `hyku_addons`, but could be injected into `Hyku::API::V1::FilesController` (https://github.com/ubiquitypress/hyku-api/blob/main/app/controllers/hyku/api/v1/files_controller.rb).  It will probably want to use `Hyrax::FileSetSearchBuilder` instead of `Hyrax::WorkSearchBuilder` to find the file set then route could be:
`/api/v1/tenant/:tenant_id/file/:id/download`

prdanelli:
@cjcolvar I've refactored as you requested. Can you review and let me know if the approach is acceptable? The examples from with the Hyrax codebase weren't super straight forward, so I was mostly digging around in the dark trying to make it work as they had. So some of the steps might not be required.

cjcolvar (approved):

PR #162, opened 2021-03-15
AH-225: Feature/work UUID from file set UUID

prdanelli:
Allow the API to hit an endpoint, provide a FileSet UUID and receive the Work UUID.

### DIFF
--- a/app/controllers/concerns/hyku_addons/files_controller_behavior.rb
+++ b/app/controllers/concerns/hyku_addons/files_controller_behavior.rb
@@ -5,18 +5,10 @@ module HykuAddons
 
     # /api/v1/tenant/:tenant_id/files/:file_set_id/work
     def work
-      render plain: file_set.parent.attributes.dig("id"), status: 200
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/hyku_addons/files_controller_behavior.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    private
+
+      def file_set
+        @file_set ||= begin
+          return unless file_presenter&.current_ability&.can?(:read, file_document.id)
+
+          FileSet.find(file_document.id)
+        end
+      end
```
cjcolvar, line 19: It looks like file set presenter has a parent method which will hit solr instead of fedora so this could be reduced to:
```suggestion
      render plain: file_presenter.parent&.id, status: 200
    end

    private
```

prdanelli: Thanks @cjcolvar thats great.

cjcolvar (approved):

PR #164, opened 2021-03-16
REPO-832: Remove DOI tab for non-admin users

prdanelli:


cjcolvar (approved):
:+1: I think these tests might be improved if the private method wasn't stubbed, but this is good enough for now.

cjcolvar (approved):

PR #165, opened 2021-03-16
Overides partial and visibility of link unless admin

BertZZ:


app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
```diff
@@ -0,0 +1,15 @@
+<li class="h5"><%= t('hyrax.admin.sidebar.repository_objects') %></li>
+  <% if can?(:manage, :collection_types) %>
```
BertZZ, line 2: Could not find a helper, open to feedback on this conditional

cjcolvar: How about `can?(:manage, Collection)`?  That would make sense to me.

cjcolvar: I think we'll also want to add this as a feature flipper. See Paul's PR for how to do this: https://github.com/ubiquitypress/hyku_addons/pull/164/files
So the logic would become "if feature enabled and can manage Collection"

BertZZ: Sure I'll take a look

app/views/hyrax/dashboard/sidebar/_menu_partials.html.erb
```diff
@@ -0,0 +1,4 @@
+<% sidebar_partials = { activity: [], configuration: [], repository_content: [], tasks: [] } %>
```
cjcolvar, line 1: Is overriding this file necessary?  I wouldn't expect to need it since it should be coming from upstream Hyrax.

cjcolvar (commented):
Looks good.  I just had one question.

app/views/hyrax/dashboard/sidebar/_menu_partials.html.erb
BertZZ: Yeah it was really annoying and why it took so long `<%= render 'hyrax/dashboard/sidebar/menu_partials', menu: menu, section: :repository_content %>` Kept telling me it couldn't find this partial

cjcolvar: Huh, that's weird.  We should make sure to come back and fix this at some point.

cjcolvar (commented):
Looks like something isn't right because two tests are failing.
rspec ./spec/features/create_collection_spec.rb:21 # Create a Collection creates a collection without creator or contributor
rspec ./spec/features/create_collection_spec.rb:38 # Create a Collection creates a collection with creator

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
```diff
@@ -0,0 +1,15 @@
+<li class="h5"><%= t('hyrax.admin.sidebar.repository_objects') %></li>
+  <% if current_user.can?(:manage, Collection) || Flipflop.enabled?(:show_collections_link) %>
```
BertZZ, line 2: Finally cracked that this needed to be or not and

cjcolvar (changes requested):
Sorry, just looking at this a bit more now.  It looks like the `_repository_content.html.erb` we should be starting from is this one: https://github.com/samvera/hyku/blob/master/app/views/hyrax/dashboard/sidebar/_repository_content.html.erb  which should remove the render call and the need for the other overridden partial.  The hyku version has the bulkrax links which we'll need.

cjcolvar (approved):
:tada: Looks good!

PR #166, opened 2021-03-16
Update hydra-head for deprecation warning fix

cjcolvar:


cjcolvar:
This version of hydra-access-controls doesn't actually have the fix.

PR #173, opened 2021-03-17
Fix file availability facet

cjcolvar:
Switch test to use PacificArticle work type because GenericWork work type was mysteriously failing for me locally.


okeen (approved):
OK!

PR #174, opened 2021-03-18
Reorder notifications onload

prdanelli:
This is kind of janky, however it seems like a good place to start. I would prefer to do this inside of the Ruby libraries, however it would cause other side affects around the application.


cjcolvar (approved):
This looks good for now.  It probably makes sense to propose the actual change in the rails code upstream in Hyrax then we can remove this override when that gets merged.

PR #177, opened 2021-03-18
Cleans up deposit forms based on Pacific feedback

BertZZ:


cjcolvar (approved):
I'm a bit confused by this since it looks like non-Pacific models have both `version` and `version_number`.  But if only `version_number` works (which I think it should) then go for it.

BertZZ:
@cjcolvar I don't think `version` is showing up on the non pacific forms either

PR #181, opened 2021-03-19
Simplify the logic in deciding the locale name

prdanelli:
@cjcolvar this might be what you were after by not using the REPO name at all and if the setting isn't set, default to the normal locale. 

I've also added a couple of request-side specs to ensure that the locale is properly set.


spec/requests/hyku_addons/multitenant_locale_controller_behavior_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  describe "when a locale name is set on an account" do
+    before do
+      account.settings["locale_name"] = "test"
+      account.save
+
+      get main_app.polymorphic_path(work)
+    end
+
+    it "appends the locale name to the languge code" do
+      expect(response.body).to include('lang="en-TEST"')
```
cjcolvar, line 43: Where is this in the body?  I would hope that `en-TEST` isn't displayed to the user at all: location url or page body.  Instead I would hope the user only ever sees `en`.

app/controllers/concerns/hyku_addons/multitenant_locale_controller_behavior.rb
```diff
@@ -23,17 +23,16 @@ def default_url_options
...
         params[:locale] || I18n.default_locale
       end
 
-      # Use the setting if its been set or fall back to the tenants name.
-      # NOTE: The settings value can be "" or nil, so presence is used to check
-      def current_tenant_name
-        current_account&.settings&.dig("locale_name").presence || current_account&.name
+      # NOTE: The settings value can be "" or nil, so use `presence` if checking present?
+      def current_locale_name
+        current_account&.settings&.dig("locale_name")
```
cjcolvar, line 35: This changes it to opt-in via the setting instead of always on by default and overridable via the setting.  I could get behind this I think.  Make it explicit.

spec/requests/hyku_addons/multitenant_locale_controller_behavior_spec.rb
prdanelli: It's not visible to the user, but part of the markup. The user only ever sees the locale in the URL (if they look there). This seemed a simple way to confirm that the locale was loaded. However, if you know if a better way, i'm more than happy to change it.

app/controllers/concerns/hyku_addons/multitenant_locale_controller_behavior.rb
prdanelli: My thinking was that there was no real need to have a custom locale set, then fall back to the default, if we are only ever going to trigger the custom locale when specific clients want to alter text around the admin. It seemed better to just use the default in that case.

cjcolvar: Looking at this again.  Do we need to restore the `.presence`?  Or does this still work if the setting is empty string `""`?

PR #190, opened 2021-03-24
Bump api version for improved pagination of approval workflow comments

cjcolvar:


cjcolvar:
This is not legal because the upgrade of `mimemagic` would require relicensing the whole stack as GPLv2.

PR #191, opened 2021-03-24
Update hyku-api

cjcolvar:
Pin `mimemagic` to yanked version of previously used gem temporarily until real solution is found by rails community.


okeen (approved):
Nice!

PR #192, opened 2021-03-24
REPO-907: DOI Autocomplete Improvements

prdanelli:


app/assets/javascripts/hyku_addons/prefill_work_form_via_doi.js
```diff
@@ -3,8 +3,16 @@ class PrefillWorkFormViaDOI {
     this.targetInputSelector = "select, input, textarea, checkbox, radio"
     this.buttonSelector = "#doi-autofill-btn"
     this.form = $(this.buttonSelector).closest("form")
+    // Object wide caceh for items in arrays
```
fradeve, line 6: `caceh`?

prdanelli: English is hard. :(

app/services/bolognese/writers/hyku_addons_work_form_fields_writer.rb
```diff
@@ -3,29 +3,92 @@
...
-          'keyword' => subjects&.pluck("subject")
+          "abstract" => write_descriptions,
+          "keyword" => subjects&.pluck("subject")
         }
       end
 
       protected
 
+        def write_funders
+          funding_references.map do |funder|
```
cjcolvar, line 39: I'm a bit concerned with composing these two together (DOI metadata fetching and ROR metadata fetching).  There could easily be another data source we want to integrate later and I don't think it makes sense to inline all of them.  I guess I was thinking at the least having the ROR api handling done in a separate service.  It would be good to figure out an interface for this.  Maybe there should be a new endpoint which has a new backing service class which orchestrates fetching all of this linked data in the right order and returning JSON for populating the deposit form.  It would be nice if it could also be used by backend processes as well like for the autopopulation in AH WP1.

prdanelli: These are all fair points. I had been focusing on getting all this working, then could start to abstract it out.

cjcolvar (approved):
Awesome work!  This autofill functionality is getting better all the time and is going to be what makes the future Advancing Hyku work be amazing! :sparkles:

PR #195, opened 2021-03-25
Hides import and export links for non admin users

BertZZ:


config/features.rb
```diff
@@ -8,9 +8,9 @@
           default: false,
           description: "The admin set to deposit works is selected from the add work modal and the relationships tab on the deposit form is hidden for depositors."
 
-  feature :show_collections_link,
+  feature :show_repository_objects_links,
           default: true,
-          description: "Shows collections tab on sidebar for non admin users"
+          description: "Shows collections, import and expot links on the sidebar for non admin users"
```
cjcolvar, line 13:
```suggestion
          description: "Shows collections, importer and exporter links on the sidebar for non admin users"
```

spec/features/create_collection_spec.rb
```diff
@@ -18,51 +18,69 @@
     login_as user
   end
 
-  it 'creates a collection without creator or contributor' do
-    visit '/dashboard'
-    click_link "Collections"
-    click_link "New Collection"
+  context "when user is admin" do
+    it 'creates a collection without creator or contributor' do
+      user.add_role(:admin)
```
cjcolvar, line 23: I think this assumes the tests run in a certain order.  I think it would be better to add this to a `let` within the current context like:
`let(:user) { create(:admin)`
Although I wouldn't think this necessary if the feature is enabled by default because then it should appear for all users with dashboard access.

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
```diff
@@ -1,22 +1,22 @@
 <%# override from bulkrax to make bulkrax a feature flipper %>
 <li class="h5"><%= t('hyrax.admin.sidebar.repository_objects') %></li>
 
-<% if current_user.can?(:manage, Collection) || Flipflop.enabled?(:show_collections_link) %>
```
cjcolvar, line 4: I have some slight concern about moving this link below works for all cases.  I think it does make sense for the works link to stay constant at the top of the list when other links are appearing/disappearing based upon role, but if the feature isn't disabled then I think this changes the upstream behavior and I'm not sure if there is a strong enough case to change that behavior..

spec/features/create_collection_spec.rb
BertZZ: Ahhh you might be right, this might be left over from when I had it off by default

app/views/hyrax/dashboard/sidebar/_repository_content.html.erb
BertZZ: It was mostly just because otherwise we would need to introduce another loop into the view and considering it is doing the exact same thing, moving the links seemed the better option

spec/features/create_collection_spec.rb
```diff
@@ -65,4 +65,19 @@
 
     expect(page).not_to have_content('Contributor')
   end
+
+  context "when user is not an admin" do
+    it "Does not display the collections link when the setting is off" do
+      user.remove_role(:admin) # Ensure user is not an admin for this test
```
cjcolvar, line 71: Instead of this, how about setting user in the top describe context to an admin user:
`let(:user) { FactoryBot.create(:admin) }`
then in this context override it:
`let(:user) { FactoryBot.create(:user) }`
I think that would be clearer then how currently the user is an admin just because of the email address which isn't very explicit.
It would also be nice to replicate this context block for `when user is an admin` just to test those cases.

cjcolvar (approved):
Looks good!  I think we can merge this now and see if we get any feedback about link order which should be easy to change if we need to.

PR #199, opened 2021-03-29
Adds official link as DOI and ensures correct order of fields on depo…

BertZZ:
…sit form

### DIFF
--- a/config/locales/en-PACIFIC.yml
+++ b/config/locales/en-PACIFIC.yml
@@ -33,8 +33,10 @@ en-PACIFIC:
         irb_status: IRB Status
         irb_number: IRB Number
         version_number: Version
+        official_link: DOI
     hints:
       defaults:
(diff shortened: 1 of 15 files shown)


prdanelli (approved):

config/locales/en-PACIFIC.yml
```diff
@@ -33,8 +33,10 @@ en-PACIFIC:
         irb_status: IRB Status
         irb_number: IRB Number
         version_number: Version
+        official_link: DOI
```
cjcolvar, line 36: @BertZZ Can you add a default translation in `en.yml` to match what was removed from `_attribute_rows.html.erb`?

BertZZ: :+1:

cjcolvar (approved):

PR #202, opened 2021-03-29
Import mode for running all background jobs on special tenant_import_queue queues

cjcolvar:


app/jobs/concerns/hyku_addons/import_mode.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+module HykuAddons
+  module ImportMode
+    extend ActiveSupport::Concern
+
+    def queue_name
+      if non_tenant_job?
```
prdanelli, line 7: Could you use a guard here to simplify it a little?

```return super if non_tenant_job?```

prdanelli (approved):

app/jobs/concerns/hyku_addons/import_mode.rb
cjcolvar: :+1: I was modelling this off of https://github.com/samvera/hyku/blob/master/lib/active_job_tenant.rb#L33-L41 but I like the guards approach.

cjcolvar: I'm kind of surprised that rubocop didn't complain about this.

PR #219, opened 2021-04-08
REPO-1007: Adds validation script to compare original and imported wo…

okeen:
…rk metadata.WIP


app/services/hyku_addons/api_work_json_service.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    ##
+    # @param [Work]
+    def initialize(account, work_id)
+      @account = account
+      @work_id = work_id
+      @service_base_path = @account.settings["bulkrax_imports_server"]
+      @import_tenant_id = '3ac7fcf1-2ed3-4e47-aebd-93f10b55fe0a'
+      @username = 'elisa.barrett@ubiquitypress.com'
+      @password = 'Ubiquity'
```
cjcolvar, line 15: These probably need to change ASAP.

cjcolvar (commented):
This is good work, but I'm still a bit afraid about merging it and running it as is.  My concern is that I don't want to scare the tenant by lots of false positives so I'm wondering if there is a way to generate the report without Bulkrax::Status entries.  Sorry for giving you the run around with this.  Let's chat about it and figure something out.

app/services/hyku_addons/api_work_json_service.rb
okeen: OMG!

app/jobs/hyku_addons/validate_importer_works_job.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+class ValidateImporterWorksJob < ApplicationJob
+  # non_tenant_job
+
+  def perform(account, importer)
+    AccountElevator.switch! account.cname
+    importer.entries.find_each do |entry|
+      next if entry.is_a?(HykuAddons::CsvAdminSetEntry)
```
cjcolvar, line 9: Does this do okay with `HykuAddons::CsvCollectionEntry` objects?
Otherwise we could do:
```suggestion
      next unless entry.is_a?(HykuAddons::CsvEntry)
```

app/services/hyku_addons/importer_validation_service.rb
```diff
@@ -0,0 +1,36 @@
...
+
+      raise ArgumentError, "You must pass a valid Account" unless @account.present?
+      raise ArgumentError, "You must pass a valid HykuAddons::Importer" unless @importer.present?
+      raise ArgumentError, "Validation can only be made against successfully imported items" unless @importer.status == "Complete"
+      valid_endpoint_params?
+    end
+
+    def validate
+      @importer.entries.find_each.map do |entry|
+        next if entry.is_a?(HykuAddons::CsvAdminSetEntry)
```
cjcolvar, line 21: Same here?

app/views/bulkrax/entries/show.html.erb
```diff
@@ -0,0 +1,66 @@
...
+          <% if record.is_a?(Collection) %>
+            <%= link_to record.class.to_s, hyrax.polymorphic_path(record) %>
+          <% else %>
+            <%= link_to record.class.to_s, main_app.polymorphic_path(record) %>
+          <% end %>
+        <% else %>
+          <strong>Item Link:</strong> No item associated with this entry or class unknown
+        <% end %>
+      <% end %>
+    </p>
```
cjcolvar, line 63: It would be nice to fix this while we're here but it is lower priority.

cjcolvar (commented):
Is the job queued anywhere?  It looks like the rake task calls `ImporterValidationService` which calls `EntryValidationService` directly.  It would be great to kick this out to the background job somethere to multithread it.

cjcolvar (approved):

okeen:
Woohoo! 👯

PR #220, opened 2021-04-09
Wp3/feature/orcid integration

prdanelli:
ORCID Integration for WP3. 

NOTE: 

This was the branch where the original code was developed and some relating to HykuAddons can be still found, however the branch with the Gem integration can be found [here](https://github.com/ubiquitypress/hyku_addons/pull/453)


cjcolvar:
Before you get too far down this road, I think it is worth considering if this feature best belongs in `hyku_addons` or somewhere else.  I think it may make sense to spin off a new `hyrax-orcid` engine because I don't think there is anything hyku specific in it and there is interest in using it apart from the monolithic hyku_addons plugin. What do you think?
@BertZZ has created a new plugin repository before for `hyrax-hirmeos` so he could help you get started.

prdanelli:
> Before you get too far down this road, I think it is worth considering if this feature best belongs in `hyku_addons` or somewhere else. I think it may make sense to spin off a new `hyrax-orcid` engine because I don't think there is anything hyku specific in it and there is interest in using it apart from the monolithic hyku_addons plugin. What do you think?
> @BertZZ has created a new plugin repository before for `hyrax-hirmeos` so he could help you get started.

@cjcolvar sounds like a good idea and keeping distinct features out of the main AH plugin would be beneficial long term. I think developing a gem in the early stages is abit of a hassle given the server restarts required, but I could use the PR to extract the functionality to a Gem once its a little further along... unless you have any workflow tips from your Hyrax DOI work?

prdanelli:
This was the first stage in writing the Gem and is no longer required.

PR #225, opened 2021-04-12
Hotfix/delegation issues

prdanelli:
Refactor Hyku Solr delegation


app/presenters/concerns/hyku_addons/presenter_delegatable.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module HykuAddons
```
prdanelli, line 3: I think this would be best as a class method that took an array of fields as an argument, rather than a concern with an array. That is the next step i think.

app/presenters/concerns/hyku_addons/work_presenter_behavior.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
```
prdanelli, line 1: This concern takes the place of the GenericWorkPresenter, which is more specially targeted to adding functionality to that Presenter, rather than being a generic concern for all work presenters.

app/views/hyrax/base/_attribute_rows.html.erb
```diff
@@ -11,7 +11,7 @@
 <%= presenter.attribute_to_html(:institution, html_dl: true) %>
 <%= presenter.attribute_to_html(:org_unit, label: "Organisational unit", html_dl: true) %>
 <%= presenter.attribute_to_html(:project_name, html_dl: true) %>
-<%= render "hyku_addons/funder/show", presenter: presenter %>
+<%= render "hyku_addons/funder/show", presenter: presenter if presenter.respond_to?(:funder) %>
```
cjcolvar, line 14: I think it may be good to eventually get these to use the `presenter.attribute_to_html` pattern that all of the other fields use.

app/views/hyku_addons/current_he_institution/_show.html.erb
```diff
@@ -1,8 +1,8 @@
-<!-- rendered from views/hyrax/base///_attribute_rows.html.erb -->
-<% array_of_hash = get_model(presenter.current_he_institution, presenter.model, 'current_he_institution', nil) %>
+<% if presenter.respond_to?(:current_he_institution) %>
```
cjcolvar, line 1: Is this guard necessary since it was also added to the render in `_attribute_rows.html.erb`

app/views/shared/_additional_citations.html.erb
```diff
@@ -0,0 +1,12 @@
+<% content_for(:gscholar_meta) do %>
+	<% if @presenter.doi.present? %>
+		<meta name="citation_doi" content="<%= @presenter.doi %>"/>
+	<% end %>
+
+	<% if @presenter.respond_to?(:isbns) && @presenter.isbns.present? %>
```
cjcolvar, line 6: This should probably get contributed back upstream.

cjcolvar (approved):
Looks good!

PR #229, opened 2021-04-14
Fix field used for collection facet so it displays label instead of id

cjcolvar:


cjcolvar:
This actually isn't needed and reindexing is what was really needed.

PR #231, opened 2021-04-14
Try to resolve 500 error with PacificArticle

prdanelli:


BertZZ (approved):

prdanelli:
@cjcolvar This is a shot in the dark, cause if this doesn't work, we'll need to add some extra logging to the API and log the data to sentry... if thats even possible.

BertZZ:
If nothing else this still needs to get done

PR #239, opened 2021-04-16
Creates work_factory_behavior module

BertZZ:


app/controllers/concerns/hyku_addons/work_factory_behavior.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
```
BertZZ, line 1: I know this isn't where this file goes but its not loading from lib so I put it here till I can figure out how to get it working in the engine from lib

BertZZ:
This PR and branch is a mess, closing and taking the parts that work to a cleaner branch

PR #244, opened 2021-04-20
REPO-1045: Add middleware actor for setting member collection id for admin set

prdanelli:


app/actors/hyku_addons/actors/member_collection_from_admin_set_actor.rb
```diff
@@ -0,0 +1,39 @@
...
+      delegate :update, to: :next_actor
+
+      protected
+
+        def ensure_collection!(env)
+          admin_set_id = env.attributes.dig("admin_set_id")
+
+          return unless admin_set_id.present?
+
+          title = AdminSet.find(admin_set_id).title.first
```
okeen, line 22: `find` could throw an exception here if it can't find the record with that admin_set_id. Would that be possible on the workflow?

app/actors/hyku_addons/actors/member_collection_from_admin_set_actor.rb
```diff
@@ -0,0 +1,39 @@
...
+          collection = Collection.where(title: title)
```
okeen, line 23: I think you can merge this line with the next. Also, `where` returns a collection proxy rather than a single instead like `find_by` would. It will work fine as long as there is a unique validation on the collection titles, but it is worth mentioning.

spec/actors/hyku_addons/actors/member_collection_from_admin_set_actor_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    allow(Flipflop).to receive(:enabled?).with(flipflop_name).and_return(true)
+  end
+
+  describe "#create" do
+    context "when the flipflop is enabled" do
+      let(:attributes) { { admin_set_id: admin_set.id } }
+
+      it "called the terminator" do
+        allow(terminator).to receive(:create).with(env_class)
+
```
okeen, line 47: rubocop will probably say something about these spaces :D

spec/actors/hyku_addons/actors/member_collection_from_admin_set_actor_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+    context "when the user is an admin" do
+      let(:attributes) { { admin_set_id: admin_set.id } }
+
+      before do
+        # TODO: Workout how to get upstream hyku factories so this isn't required and we can use `build(:admin)`
+        user.add_role(:admin)
+      end
+
+      it "calls the termintor" do
```
okeen, line 98: Small typo: terminAtor

spec/actors/hyku_addons/actors/member_collection_from_admin_set_actor_spec.rb
```diff
@@ -0,0 +1,165 @@
...
+        allow(terminator).to receive(:create)
```
okeen, line 99: I normally set up the spies / stubs into the `before` block, as it is part of the test set up. WDYT?

okeen (changes requested):
Let's wait for Rubocop :D

app/actors/hyku_addons/actors/member_collection_from_admin_set_actor.rb
prdanelli: I did try both `find_by` isn't in use because its `ActiveFedora` not `ActiveRecord` and using `where` stops me from having to deal with the `nil` errors if an `AdminSet` isn't found.

app/actors/hyku_addons/actors/member_collection_from_admin_set_actor.rb
prdanelli: So, because the admin_set is only set from the available options, i think this is ok for now. If the user messes with the URL params, thats on them.

spec/actors/hyku_addons/actors/member_collection_from_admin_set_actor_spec.rb
okeen: Sorry, I meant extra new lines.

PR #251, opened 2021-04-26
Remove now redundant code

prdanelli:
Previously there were issues with crossref DOI's "10.7554/eLife.67932" and "10.7554/elife.65703", which both seemed to mess up their funding information. This seems to have been fixed elsewhere, so i've removed the overload method and added specs to ensure that one of these doi's funder information is checked for regressions.


cjcolvar (approved):
Looks great!  Thanks for cycling back and cleaning this up!

PR #261, opened 2021-04-29
Installs hyrax-hirmeos and adds work factory overide

BertZZ:
Do it again, but better this time!


cjcolvar (approved):
This looks great!
Now that `hyrax-hirmeos` is in the Gemfile.lock we'll need to update the version in `hyku_addons` for it to get into the built container image like `hyku-api`.

PR #268, opened 2021-05-07
Switch settings on account switch

cjcolvar:
Reload rubyconfig Settings with any tenant specified settings or
settings overrides with the following precedence (lowest to highest):
settings.yml
settings/{environment}.yml
settings/{environment}.local.yml
settings/{environment}-{TENANT_LOCALE_NAME}.yml
current_account.settings
ENV["SETTINGS__#{SETTING_NAME}"]
ENV["SETTINGS__#{TENANT_LOCALE_NAME}__#{SETTING_NAME}"]

After this is in place we should probably centralize all reading of settings to `Settings` (with the exception of `locale_name` used here to load the settings).  So `current_account.allow_signups` would become `Settings.allow_signups`.

Closes https://ubiquitypress.atlassian.net/browse/REPO-1115 and https://ubiquitypress.atlassian.net/browse/REPO-1116

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -68,6 +68,7 @@ PATH
   specs:
     hyku_addons (0.1.0)
       blacklight_oai_provider (~> 6.1)
+      config (>= 3.0)
       google-cloud-storage (~> 1.31)
       hyrax (~> 2.8)
       hyrax-doi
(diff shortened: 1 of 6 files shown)


cjcolvar:
Still need to add tests.

cjcolvar:
I manually tested creating two tenants locally and manually added a GA id to one of them.  This appeared to work successfully as the tenant with the GA Id had GA JS with the id in the `<head>` but the tenant without did not have that JS.

prdanelli (approved):

PR #270, opened 2021-05-10
Create Redlands Article

BertZZ:


app/views/records/edit_fields/_creator.html.erb
```diff
@@ -34,7 +34,7 @@
 		<% end %>
 
 		<% field_options = { field_slug: :creator_isni, required: false, field_group: :creator, template: template, hash: hash } %>
-		<%= render "records/edit_fields/attribution_partials/text_field", field_options %>
+		<%= render "records/edit_fields/attribution_partials/text_field", field_options unless template.include? "redlands" %>
```
prdanelli, line 37: I feel like this is the second time that we have included a line like this, once for Pacific and once here. 

Instead of `unless template.include? "redlands"`, could you include a Flipflop to hide the isni for the tenant and that could be used on pacific as well.

app/views/records/edit_fields/_creator_personal.html.erb
```diff
@@ -1,15 +1,16 @@
 <div data-toggleable-group="<%= type %>">
 	<% creator_inst_rel_options = HykuAddons::InstitutionalRelationshipService.new(model: f.object.model.class).select_active_options %>
+	<% creator_role_options = HykuAddons::RoleService.new(model: f.object.model.class).select_active_options %>
 	<% personal_fields = [
 		{ field_type: :text, field_slug: :creator_family_name, field_args: { data: { required: f.object.required?(:creator), required_group: :creator_name, on_blur: :toggle_required_group } } },
 		{ field_type: :text, field_slug: :creator_given_name, field_args: { data: { required: f.object.required?(:creator), required_group: :creator_name, on_blur: :toggle_required_group } } },
 		{ field_type: :text, field_slug: :creator_orcid },
 		{ field_type: :select, field_slug: :creator_institutional_relationship, select_options: creator_inst_rel_options, field_args: { multiple: true } }] %>
 	<% personal_fields = add_pacific_creator_personal_fields(personal_fields) if template.include? "pacific" %>
```
prdanelli, line 9: The fact that we have 3 clients and all these require a specific method here, is a pretty big code smell that we need to think of a better way to do this, so that in the future, if we have a thousand customers, we dont have a thouand methods here.  

`add_pacific_creator_personal_fields`, `add_uva_creator_personal_fields`, `add_redlands_creator_personal_fields`

app/views/redlands/related_identifier/_edit_array_hash_form.html.erb
```diff
@@ -0,0 +1,32 @@
+<% template = @curation_concern.class.to_s.underscore %>
+
+<% array_of_hash.each_with_index do |hash, index| %>
+
```
prdanelli, line 4: This line break isn't required.

app/views/redlands/related_identifier/_new_form.html.erb
```diff
@@ -0,0 +1,30 @@
+<% template = f.object.model.class.to_s.underscore  %>
+
+<div class="ubiquity-meta-related-identifier" >
```
prdanelli, line 3: The spac before the `>` isnt required.

prdanelli (changes requested):

app/views/records/edit_fields/_creator_personal.html.erb
prdanelli: @cjcolvar tagging you in here too. We need a different approach.

PR #277, opened 2021-05-13
Fix Date error for incorrect DOI 

prdanelli:
Fix an issue when an incorrect DOI is entered and the date function tried to run on a nil object

### DIFF
--- a/app/services/bolognese/writers/hyku_addons_work_form_fields_writer.rb
+++ b/app/services/bolognese/writers/hyku_addons_work_form_fields_writer.rb
@@ -466,8 +466,6 @@ module Bolognese
 
         # `dates` is an array of hashes containing a string date and named dateType
         def collect_date(type)
-          return unless dates.present?
-
           dates.find { |hash| hash["dateType"] == type }&.dig("date")
         end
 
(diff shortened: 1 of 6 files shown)


app/services/bolognese/writers/hyku_addons_work_form_fields_writer.rb
```diff
@@ -466,6 +466,8 @@ def hash_from_date(field, dates)
 
         # `dates` is an array of hashes containing a string date and named dateType
         def collect_date(type)
+          return unless dates.present?
```
okeen, line 469: Just a sanity check. Did you want to check that the array is present or that it has at least one item?

app/services/bolognese/writers/hyku_addons_work_form_fields_writer.rb
```diff
@@ -498,13 +500,11 @@ def ensure_creator_from_editor!
...
         def process_editor_contributors!
-          editors = Proc.new { |item| item["contributor_contributor_type"] == "Editor" }
+          editors = proc { |item| item["contributor_contributor_type"] == "Editor" }
 
           # If we do not have editors, they might be missing from the fields for this work type.
           # This is so that we can reliably use them later on in the callbacks chain
-          if @form_data["editor"].blank? && @form_data["contributor"].any?(&editors)
-            @form_data["editor"] = @form_data["contributor"].select(&editors)
-          end
+          @form_data["editor"] = @form_data["contributor"].select(&editors) if @form_data["editor"].blank? && @form_data["contributor"].any?(&editors)
```
okeen, line 507: Another sanity check. An empty array of `editors` would pass the second check of the if expression. Is that what you want?

okeen (approved):

PR #279, opened 2021-05-14
REPO-1126 - Task Master Integration

prdanelli:


config/features.rb
```diff
@@ -19,4 +19,8 @@
   feature :import_mode,
           default: false,
           description: "Import mode - Background jobs are run on specially named queues"
+
```
fradeve, line 22: I am not sure what this `feature` is for, but this is going to become an integral part of our repository setup and will be enabled for every repository we run -- ever. So won't ever need to disable it.
As far as I am concerned, this can be removed :)

prdanelli: @fradeve I was thinking more of other people using the HA codebase that didn't have this setup in mind. as this feature is very specific to Ubiquity.

cjcolvar: :+1: This should get pulled out to the UP specific engine whenever we make that or it could get pulled out into its own specific little gem.

prdanelli: I was thinking the same thing. It would be best in its own Gem probably, but i'll carry on developing inside HA for now.

app/actors/hyku_addons/actors/member_collection_from_admin_set_actor.rb
```diff
@@ -21,9 +21,12 @@ def ensure_collection!(env)
 
           title = AdminSet.find(admin_set_id)&.title&.first
           collection = Collection.where(title: title).first if title.present?
+
+          return unless collection.present?
```
cjcolvar, line 25: :+1:

Gemfile.lock
```diff
@@ -508,6 +508,7 @@ GEM
       google-cloud-core (~> 1.2)
       googleauth (~> 0.9)
       mini_mime (~> 1.0)
+    google-protobuf (3.17.0)
```
cjcolvar, line 511: Was there a change to the Gemfile that brought these in?  Just checking that everything got committed that should be.

app/services/hyku_addons/task_master/publish_service.rb
```diff
@@ -0,0 +1,73 @@
...
+                        end
+
+                        Google::Cloud::PubSub.new
+                      end
+        end
+
+        def pubsub_credentials
+          raise KeyError, "Service environment variable is not set" unless ENV["PUBSUB_SERVICEACCOUNT_KEY"].present?
+
+          @pubsub_credentials ||= JSON.parse(ENV["PUBSUB_SERVICEACCOUNT_KEY"])
```
cjcolvar, line 61: I wonder if we should pipe this through `Settings` as well.  This isn't vital right now just think it might be good to do eventually.

config/features.rb
```diff
@@ -19,4 +19,8 @@
...
+  feature :task_master,
+          default: false,
```
cjcolvar, line 24: What do you think about setting this based upon the ENV?  This way we don't have to remember to turn it on for every tenant.
```suggestion
          default: ENV["PUBSUB_SERVICEACCOUNT_KEY"].present?,
```

spec/factories/works.rb
```diff
@@ -51,4 +51,23 @@
     creator { [''] }
     contributor { [''] }
   end
+
+  factory :task_master_work, parent: :work do
```
cjcolvar, line 55: Do you think we could come up with a more generic name for this factory?

spec/actors/hyku_addons/actors/task_master_work_actor_spec.rb
```diff
@@ -0,0 +1,157 @@
+# frozen_string_literal: true
```
cjcolvar, line 1: Looks like this spec could be removed in favor of `spec/actors/hyku_addons/actors/task_master/work_actor_spec.rb` above.

cjcolvar (approved):
This looks great!  I have a few questions/suggestions but nothing that I think is really blocking this from being merged.

fradeve:
@prdanelli please let's remember to squash the commits before merging... the history on this branch is getting bigger :)

fradeve:
@prdanelli very very minor, but the name of the service is `taskmaster` rather than `Task Master` or `task_master`

prdanelli:
> @prdanelli very very minor, but the name of the service is `taskmaster` rather than `Task Master` or `task_master`

@fradeve so it would be `Taskmaster` if we are naming classes?

cjcolvar (approved):

PR #290, opened 2021-05-20
Send JSON string instead of hash to Taskmaster publish job

cjcolvar:


prdanelli:
This change is covered in a more recent fix. Thank you though.

PR #296, opened 2021-05-22
REPO-1104: Added note column and tab for adding notes in deposit form

ashikajith:
Added notes tab in deposit form and the respective model actors for processing the note content for each model objects

Screenshot from 2021-05-23 23-46-00

### DIFF
--- a/app/actors/hyku_addons/actors/note_field_actor.rb
+++ b/app/actors/hyku_addons/actors/note_field_actor.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+module HykuAddons
+  module Actors
+    class NoteFieldActor < Hyrax::Actors::AbstractActor
+      attr_accessor :email
+
+      def create(env)
+        serialize_note_field(env) && next_actor.create(env)
+      end
+
+      def update(env)
+        serialize_note_field(env) && next_actor.update(env)
+      end
+
+      private
+
+        def serialize_note_field(env)
+          if env.attributes[:note].present?
+            @email = env.user.email
+            if env.curation_concern.note.present?
+              new_note_value = env.curation_concern.note + Array(process_note_hash(env.attributes[:note]).to_json)
+            else
+              new_note_value = Array(process_note_hash(env.attributes[:note]).to_json)
+            end
+            env.attributes[:note] = new_note_value
+          end
+        end
+
+        def process_note_hash(note)
+          { email: email, timestamp: Time.zone.now.to_s, note: note }
+        end
+    end
+  end
+end
--- a/app/actors/hyku_addons/actors/note_field_actor.rb
+++ b/app/actors/hyku_addons/actors/note_field_actor.rb
@@ -17,12 +17,14 @@ module HykuAddons
         def serialize_note_field(env)
           if env.attributes[:note].present?
             @email = env.user.email
-            if env.curation_concern.note.present?
-              new_note_value = env.curation_concern.note + Array(process_note_hash(env.attributes[:note]).to_json)
-            else
-              new_note_value = Array(process_note_hash(env.attributes[:note]).to_json)
-            end
-            env.attributes[:note] = new_note_value
+            env.attributes[:note] =
+              if env.curation_concern.note.present?
+                env.curation_concern.note + Array(process_note_hash(env.attributes[:note]).to_json)
+              else
+                Array(process_note_hash(env.attributes[:note]).to_json)
+              end
+          else
+            env.attributes[:note] = []
           end
         end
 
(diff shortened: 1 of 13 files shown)


cjcolvar (changes requested):
This is looking really good!

The only part that I think needs some small changes is that storing the notes as separate JSON blobs for each note and how Fedora/RDF work is that there is no guaranteed order so each time you load the page the notes might appear in a different order and not chonologically.  I think either the notes will need to be sorted by timestamp in the rendering (maybe in the helper?) or they need to be stored sorted as a single JSON blob (parsing the JSON and appending/prepending to the array before reserializing it as JSON in the actor).  I think either approach would be fine.  My guess is that the notes were displayed oldest to newest before and that's what we want to do again.  What do you think?

spec/actors/hyku_addons/actors/note_field_actor_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+      it "call serialize_note_field method while creating work" do
+        middleware.create(env)
+        expect(middleware).to have_received(:serialize_note_field).with(env)
+      end
+
+      it "returns an ActiveTriples::Relation class for note attribute" do
+        middleware.create(env)
+        expect(work.note.class).to eq(ActiveTriples::Relation)
+      end
```
cjcolvar, line 43: I think it would be more helpful to see that the note got serialized as JSON and persisted on the work.  This might look like:
```
it "adds the note" do
  expect(work.note).to be_blank
  middleware.create(env)
  expect(work.note).to be_present
  expect(JSON.parse(work.note.first)).to include(email: user.email, timestamp: String, note: note)
end
```

spec/actors/hyku_addons/actors/note_field_actor_spec.rb
```diff
@@ -0,0 +1,63 @@
...
+
+    context "when the note attribute is present" do
+      before do
+        allow(terminator).to receive(:update).with(env)
+      end
+
+      it "call the update method " do
+        middleware.update(env)
+        expect(terminator).to have_received(:update).with(env_class)
+      end
```
cjcolvar, line 60: How about rewriting this one to be like my suggestion above but be adding a new note to a work that already has a note then checking that both notes are present on the saved work?

cjcolvar (commented):
Looking good!  I just had a few suggestions mostly about the specs.

app/actors/hyku_addons/actors/note_field_actor.rb
```diff
@@ -17,12 +17,14 @@ def update(env)
...
-            end
-            env.attributes[:note] = new_note_value
+            env.attributes[:note] =
+              if env.curation_concern.note.present?
+                env.curation_concern.note + Array(process_note_hash(env.attributes[:note]).to_json)
+              else
+                Array(process_note_hash(env.attributes[:note]).to_json)
+              end
+          else
+            env.attributes[:note] = []
```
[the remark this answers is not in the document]
ashikajith: in the case of nil, the deposit form was throwing an error stating the value is not scalar. It was resolved when I passed the blank array.

spec/helpers/notes_tab_form_helper_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+    allow(Flipflop).to receive(:enabled?).with(:simplified_admin_set_selection).and_return(false)
+    allow(Flipflop).to receive(:enabled?).with(:notes_tab_form).and_return(false)
+    allow(controller).to receive(:current_user) { user }
+  end
+
+  describe "#form_tabs_for" do
+    context "when the feature is enabled" do
+      before do
+        allow(Flipflop).to receive(:enabled?).with(:notes_tab_form).and_return(true)
+        allow(helper).to receive(:form_tabs_for).with(form: form).and_return(['notes'])
```
cjcolvar, line 30: I think this stubbing masks the true implementation.  You didn't have this before and it looked good to me.  Was there some test failure that led you to add this line?

cjcolvar (commented):
This is looking really close!  The actor spec looks good now.

cjcolvar:
It looks like you'll need to resolve some conflicts in `WorkForm`.

spec/helpers/notes_tab_form_helper_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      it "shows the notes tab" do
+        expect(helper.form_tabs_for(form: form)).to include("notes")
+      end
+    end
+
+    context "when the feature is disabled" do
+      before do
+        allow(Flipflop).to receive(:enabled?).with(:notes_tab_form).and_return(false)
+        allow(helper).to receive(:form_tabs_for).with(form: form).and_return(['relationships'])
```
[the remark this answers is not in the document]
ashikajith: I have fixed the stubbing for the `notes_tab_form_helper_spec`. I had issues while running the spec, turns out I enabled the wrong feature key. Now it has been fixed in the latest commit https://github.com/ubiquitypress/hyku_addons/pull/296/commits/2bbe73beafded45effeb4c65bbe53d9cfa381b45

cjcolvar (approved):

PR #299, opened 2021-05-25
Add support for Creator/contributor JSON content

prdanelli:
Currently the Bolognese GenericWorkReader uses the `display` for creator and contributor. This is preventing me from getting data out of the reader to use when exporting in the ORCID XML writer. The only writer that currently uses the generic work reader is the RISWriter, so that has been updated and the tests fixed.

### DIFF
--- a/app/services/bolognese/readers/base_work_reader.rb
+++ b/app/services/bolognese/readers/base_work_reader.rb
@@ -37,7 +37,8 @@ module Bolognese
       def self.nested_attributes
         {
           "container" => %w[volume issue firstPage lastPage pagination]
-        } end
+        }
+      end
 
       # An array of methods that should be called after the inital attributes have been collected.
       # These methods should modify the `@reader_attributes` variable directly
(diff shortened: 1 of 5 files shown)


prdanelli:
Failing specs are CI issues, they all pass locally.

app/services/bolognese/readers/base_work_reader.rb
```diff
@@ -67,18 +67,44 @@ def read_work(string: nil, **options)
...
+
+          transformed.compact
+
+        rescue JSON::ParserError
+          json
+        end
+
         def read_creator
-          return unless (value = @meta.fetch('creator_display', @meta.dig('creator'))).present?
+          return unless (value = @meta.fetch('creator', @meta.dig('creator_display'))).present?
```
okeen, line 92: Can you replace all the `'` for `"` ?

okeen (approved):
Looks great, mate!

app/services/bolognese/readers/base_work_reader.rb
prdanelli: Good spot. I did them all.

PR #306, opened 2021-05-26
Use both id and source_identifier for imports and exports

cjcolvar:
If you want to import something that already has a UUID then put it in
both id and source_identifier.  If you want Hyku to mint an id for you
then only put an id in source_identifier.  Exports should contain both
id and source_identifier.

TODO: Add tests that highlight this change, test import with id (only source_identifier), and export->import roundtrip editing


cjcolvar:
I think round-trip editing is a different feature so I didn't add any tests for it.  I did add one more test for the case when Hyku mints a new id an imported work.

okeen (approved):
Looks great to me!

PR #307, opened 2021-05-27
Remove override methods and update Bolognese Gem

prdanelli:


hyku_addons.gemspec
```diff
@@ -39,6 +39,7 @@ Gem::Specification.new do |spec|
   spec.add_dependency 'maremma', '< 4.8'
   spec.add_dependency 'postrank-uri', '>= 1.0.24'
   spec.add_dependency 'public_suffix', '~> 2.0.2'
+  spec.add_dependency 'bolognese', '~> 1.9.7'
```
cjcolvar, line 42: How about this instead since it should allow any 1.x greater than or equal 1.9.7?
```suggestion
  spec.add_dependency 'bolognese', '~> 1.9', '>= 1.9.7'
```

cjcolvar (commented):
Yay! :tada:
One suggestion about the gemspec change that I think would be good.

hyku_addons.gemspec
prdanelli: I think the greater than option would be great, otherwise our specs will fail - i think.

cjcolvar (approved):

PR #311, opened 2021-06-02
Repo 1159/identifier actor

BertZZ:
Creates a Related Identifier Actor to add PMID and HasMetadata to any related Identifier

### DIFF
--- a/app/actors/hyku_addons/actors/related_identifier_actor.rb
+++ b/app/actors/hyku_addons/actors/related_identifier_actor.rb
@@ -2,7 +2,6 @@
 module HykuAddons
   module Actors
     class RelatedIdentifierActor < Hyrax::Actors::AbstractActor
-
       def create(env)
         add_related_identifier_data(env) && next_actor.create(env)
       end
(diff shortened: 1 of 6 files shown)


app/actors/hyku_addons/actors/related_identifier_actor.rb
```diff
@@ -0,0 +1,29 @@
...
+      def update(env)
+        add_related_identifier_data(env) && next_actor.update(env)
+      end
+
+      private
+
+        def add_related_identifier_data(env)
+          return unless env.curation_concern.class.to_s.include? "Redlands"
+          identifier_hash = JSON.parse(env.attributes[:related_identifier].first).first if env.attributes[:related_identifier]
+          env.attributes[:related_identifier] = [add_related_metadata(identifier_hash)]
```
okeen, line 18: I like using `Array.wrap` when I try to ensure I convert something to an array because 1) Explicitly tells what it is doing 2) If what I am wrapping is already an array, it doesn't include it in yet another array:
```suggestion
          env.attributes[:related_identifier] = Array.wrap(add_related_metadata(identifier_hash))
```

app/actors/hyku_addons/actors/related_identifier_actor.rb
```diff
@@ -0,0 +1,29 @@
...
+        end
+
+        def add_related_metadata(identifier_hash)
+          return unless identifier_hash["related_identifier"].present?
+          identifier_hash["related_identifier_type"] = "PMID"
+          identifier_hash["relation_type"] = "HasMetadata"
+          [identifier_hash].to_json
```
okeen, line 25: Maybe it's what you want to do, but you are returning an array here too. `env.attributes[:related_identifier]` would be an array of arrays in the assignation above.

app/views/redlands/related_identifier/_edit_array_hash_form.html.erb
```diff
@@ -8,17 +8,8 @@
     <%= text_field_tag "#{template}[related_identifier][][related_identifier]", hash.dig("related_identifier"),
                        class: "#{template}_related_identifier form-control multi-text-field multi_value related_identifier",
                        name: "#{template}[related_identifier][][related_identifier]",
-                        placeholder: t("simple_form.placeholders.defaults.related_identifier")
+                       placeholder: t("simple_form.placeholders.defaults.related_identifier")
```
okeen, line 11: This is just a suggestion, but I like creating helpers to define these inputs into Helpers and keep the views clean and readable:

`def related_identifier_input_for(template, value)`
`    text_field_tag "#{template}[related_identifier][][related_identifier]", hash.dig("related_identifier"), ...`
`end`

app/actors/hyku_addons/actors/related_identifier_actor.rb
BertZZ: Yeah the structure needs to be `["[{\"related_identifier\":\"\",\"related_identifier_type\":\"PMID\",\"relation_type\":\"HasMetadata\"}]"]`

okeen (approved):
Looks good!

PR #313, opened 2021-06-03
Hotfix/on edit grouped require issue redux

prdanelli:
When editing an existing work (on any work type), removing a legacy multiple field would trigger the form validate and find hidden required fields on the "required_group" inputs. 

The after register hook, wa running `on-blur`, which seemed to be reading required attributes, even if the group was hidden. I can't see any noticeable issue caused by removing this additional action and there is nothing from the original commit that might lead me to understand why I added it and what issue it originally solved.


okeen (approved):
Looks good, mate!

PR #317, opened 2021-06-07
Changes label in Chapter worktype

BertZZ:
Changes the Edition label in the Chapter and Book Section worktype to Book Edition

### DIFF
--- a/config/locales/redlands_chapters_and_book_section.en.yml
+++ b/config/locales/redlands_chapters_and_book_section.en.yml
@@ -1,4 +1,4 @@
-en:
+en-REDLANDS:
   hyrax:
     icons:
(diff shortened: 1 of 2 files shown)


config/locales/redlands_chapters_and_book_section.en.yml
```diff
@@ -1,8 +1,12 @@
-en:
+en-REDLANDS:
```
cjcolvar, line 1: I think this should stay as `en` since it is for the work type and not just the `REDLANDS` locale.  If it is supposed to only be for the `REDLANDS` locale then I think you should make a new file which matches the convention (`redlands_chapters_and_book_section.en-REDLANDS.yml`) and make the change there instead.

cjcolvar (approved):

PR #322, opened 2021-06-08
File visibility can be set in bulkrax imports

cjcolvar:
CSVs for import have had a single column for files which allowed
multiple files by separating filenames with pipes.  This commit
preserves that functionality while allowing splitting this across many
columns with additional columns for visibility similar to json fields.
For example, a CSV can now have columns file_1, file_visibility_1,
file_2, file_3, file_visibility_3. See the newly added CSV fixture for
a template.  These columns will pass into the Hyrax actor stack and will
be applied through changes made to AttachFilesToWorkJob and the BaseActor.
This is currently done with overrides which can be removed once
https://github.com/samvera/hyrax/pull/4992 is merged and released in the
2.x Hyrax series.


app/models/hyku_addons/csv_entry.rb
```diff
@@ -33,6 +34,22 @@ def build_metadata
       parsed_metadata
     end
 
+    def add_file_subfields
+      file_metadata = []
+      raw_metadata.select { |k, _v| k =~ /^file_((?<subfield>.+)_)?(?<index>\d+)$/ }.each do |k, v|
+        match = k.match(/^file_((?<subfield>.+)_)?(?<index>\d+)$/)
+        file_metadata[match[:index].to_i - 1] ||= {}
```
okeen, line 41: I think this will add some clarity to this sentence and the next ones while also avoiding some code repetition.
```suggestion
        file_index = match[:index].to_i - 1
        file_metadata[file_index] ||= {}
```

app/models/hyku_addons/csv_entry.rb
```diff
@@ -33,6 +34,22 @@ def build_metadata
...
+        if match[:subfield].present?
```
okeen, line 42: I personally use this structure when I want to use the result of the evaluation of a conditional:
```suggestion
        if (subfield = match[:subfield].presence)
            file_metadata[file_index][subfield] = v
```

okeen (commented):
Looks good! Just a couple of minor code changes. Both comments affect multiple lines so you will probably don't want to commit my suggestion and make the changes on your local computer instead.

PR #327, opened 2021-06-09
Sends human readable labels to the API

BertZZ:


app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -53,7 +53,8 @@ json.issn work.try(:issn)
 json.issue work.try(:issue)
 json.journal_title work.try(:journal_title)
 json.keywords work.keyword
-json.language work.language
+language_options = HykuAddons::LanguageService.new.select_all_options.to_h
+json.language work.language.map { |id| language_options.key(id).to_s } if work.language.present?
```
cjcolvar, line 57: I think we'll need to send the model along to ensure we have the correct vocabulary.  There is also a `label` method available on the service which we could use here.  How about this?
```suggestion
if work.language.present?
  language_service = HykuAddons::LanguageService.new(model: work.class)
  work.language.map { |id| language_service.label(id) rescue nil }.compact
  json.language languages
end
```

BertZZ: Did not know about the label method, we don't send the model on the Licence service line below, but maybe we should??

cjcolvar (approved):

PR #330, opened 2021-06-10
REPO-1160 - Feature/multiple text fields

prdanelli:
Refactoring cloneable and toggleable js classes to allow us to clone individual elements, not just larger groups. 
image


app/forms/concerns/hyku_addons/person_or_organization_form_behavior.rb
```diff
@@ -27,7 +27,7 @@ def creator_fields
           creator: [
             :creator_organization_name, :creator_given_name, :creator_middle_name, :creator_family_name,
             :creator_name_type, :creator_orcid, :creator_isni, :creator_ror, :creator_grid, :creator_wikidata,
-            :creator_suffix, :creator_institution, creator_institutional_relationship: []
+            :creator_suffix, creator_institution: [], creator_role: [], creator_institutional_relationship: []
```
prdanelli, line 30: Not sure about this bit as it is site wide, and would make these params arrays for every work type that uses them.

BertZZ (approved):

PR #331, opened 2021-06-10
Add host file setup description to readme

prdanelli:


README.md
```diff
@@ -85,6 +85,27 @@ docker-compose exec web bundle exec rails g hyku_addons:install
 CFLAGS=-Wno-error=format-overflow  gem install zookeeper -v '1.4.11' --source 'https://rubygems.org/'
 ```
 
+### Dory / Host file
+
+Dory can be used to automatically configure your local development environment so that you can use local subdomains, however you may experience issues.
```
cjcolvar, line 90: What about adding this so there is a little bit more guidance for using dory (albeit not much and buried in a link).
```suggestion
Dory can be used to automatically configure your local development environment so that you can use local subdomains, however you may experience issues. (See https://github.com/samvera/hyku/#dory)
```

cjcolvar (approved):

PR #339, opened 2021-06-11
Fix the buttons as they are UGLY

prdanelli:
Match the bootstrap styles and turn the buttons from this... 
image

To this... 
image


prdanelli:
Everyone likes it.

PR #340, opened 2021-06-11
Export file visibility and embargo fields

cjcolvar:
We should probably also add support for leases here and test in both import and export, but that could come in another pull request.


okeen (approved):
Looks very good!

PR #347, opened 2021-06-15
Model, form, and indexer mixins to load configuration from metadata yaml

cjcolvar:


cjcolvar:
Closed in favor of #372

PR #354, opened 2021-06-17
REPO-1183: Added file set visibility in deposit forms

ashikajith:
Added File Set visibility in the deposit form UI
Screenshot from 2021-06-17 22-50-22

### DIFF
--- a/app/views/hyrax/uploads/_js_templates.html.erb
+++ b/app/views/hyrax/uploads/_js_templates.html.erb
@@ -49,6 +49,24 @@
       }
     });
   }
+
+  // For Disabling the File Set Visibility Options based on the Admin Set Configuration
+  // Function should be loaded before the table row loads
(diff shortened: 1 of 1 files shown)


app/views/hyrax/uploads/_js_templates.html.erb
```diff
@@ -169,6 +187,8 @@
             </button>
         </td>
     </tr>
+  {% var RecentFileId = file.id; %}
+  {% setTimeout(function(){ setVisiblityOptions(RecentFileId) }, 300); %}
```
ashikajith, line 191: Adding timeout to render the Html via javascript.

Due to the timeout, the `file.id` value was getting a null value. Hence I ended up saving it to a variable and calling the function using the value of the var.

cjcolvar (approved):
This is working great from my testing!

PR #357, opened 2021-06-21
Fixes issues in the latest version

BertZZ:


app/views/anschutz_works/edit_fields/_add_info.html.erb
```diff
@@ -1 +1 @@
-<%= f.input :add_info, as: :multi_value, input_html: { rows: 7, cols: 80, type: 'textarea' }, required: f.object.required?(key) %>
```
BertZZ, line 1: This change is needed but is throwing an error [](https://sentry.io/organizations/ubiquity-press-ltd/issues/2468560766/?project=5755215)

cjcolvar: The issue could be coming from `add_info` defined as `multiple: false` in WorkBase then redefined as `multiple: true` in `AnschutzWork`.  You might need to take it out of `WorkBase` and into each of the work types.

cjcolvar (approved):

PR #358, opened 2021-06-22
Attempt to fix `allow_signups` not found on Account

prdanelli:
First attempt at fixing the worlds most annoying bug, that only shows its self seemingly at random on local development environments and required a server restart to login. 

@cjcolvar do you have thoughts on this?

### DIFF
--- a/app/views/devise/shared/_links.html.erb
+++ b/app/views/devise/shared/_links.html.erb
@@ -3,7 +3,7 @@
 <% end -%>
 
 <%- if devise_mapping.registerable? && controller_name != 'registrations' %>
-  <%= !Account.global_tenant? && current_account.allow_signup == "true" ? (link_to t(".sign_up"), new_registration_path(resource_name)) : ' ' %><br />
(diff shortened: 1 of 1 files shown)


app/views/devise/shared/_links.html.erb
```diff
@@ -3,7 +3,7 @@
 <% end -%>
 
 <%- if devise_mapping.registerable? && controller_name != 'registrations' %>
-  <%= !Account.global_tenant? && current_account.allow_signup == "true" ? (link_to t(".sign_up"), new_registration_path(resource_name)) : ' ' %><br />
+  <%= !Account.global_tenant? && current_account.respond_to?(:allow_signup) && current_account.allow_signup == "true" ? (link_to t(".sign_up"), new_registration_path(resource_name)) : ' ' %><br />
```
cjcolvar, line 6: This looks fine to me.  Do you think we should use `try` instead to make it a little more concise?
```suggestion
  <%= !Account.global_tenant? && current_account.try?(:allow_signup) == "true" ? (link_to t(".sign_up"), new_registration_path(resource_name)) : ' ' %><br />
```

prdanelli: Yea, i always forget `try` is a thing now that `&` is around. If this works the same, i'm more than happy with that.

cjcolvar (approved):

PR #366, opened 2021-06-24
Update readme migrations guide.

prdanelli:


cjcolvar (approved):
I've never manually updated the version in the migration table, but whatever makes things work.

PR #381, opened 2021-07-01
Hyku Addons changes to Hyrax-DOI

prdanelli:
# Issues

1. initially my first work gave an error but that did not happen with my second work
2. if editing a work that was initially saved without DOI minting selected, and then selecting DOI minting options and saving, this does not mint a DOI (it should).
3. the JSON fields are not being translated to DataCite's format - esp Creator - so DataCite has something like this, which is wrong:

# Conclusions

1. Seems to be a configuration issues as its not replicatable.  
2. Outstanding
3. By overriding the Bolognese writer methods for creator/contrbutor, we need to be sending the correct data.

### DIFF
--- a/app/services/bolognese/writers/ris_writer_behavior.rb
+++ b/app/services/bolognese/writers/ris_writer_behavior.rb
@@ -51,7 +51,7 @@ module Bolognese
             "T1" => parse_attributes(titles, content: "title", first: true),
             "T2" => secondary_titles,
             "DO" => doi,
-            "AU" => to_ris(creators),
+            "AU" => to_ris(meta.dig("creators")), # Don't use `creators` method as its memoized and might not update
             "ED" => to_ris(meta.dig("editor")),
             "AB" => parse_attributes(descriptions, content: "description", first: true),
             "KW" => Array.wrap(subjects).map { |k| parse_attributes(k, content: "subject", first: true) }.presence,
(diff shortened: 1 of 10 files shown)


cjcolvar (commented):
Looks good enough to me.

app/services/bolognese/writers/ris_writer_behavior.rb
```diff
@@ -74,36 +74,38 @@ def ris
...
-          Array.wrap(parse_attributes(meta["alt_title"])) + Array.wrap(parse_attributes(meta["book_title"]))
-        end
+          # Expand nested arrays, remove any blank entries and expand into a RIS formatted string
+          def expand_nested_and_prepare(hash)
+            hash
+              .compact
+              .map do |k, v|
+                if v.is_a?(Array)
+                  # `dup` string for different object ID allowing duplicate keys in compare_by_identity hash
+                  v.map { |vi| "#{k.dup}  - #{vi}" if vi.present? }.compact.join(RIS_DELIMITER)
```
okeen, line 86: There are two spaces in the string before the "-". The same happens two lines below.

Just a sanity check.

app/services/hyku_addons/bolognese/json_fields_reader.rb
```diff
@@ -0,0 +1,32 @@
...
+
+          t["nameIdentifier"] = {
+            "nameIdentifierScheme" => "orcid",
+            "__content__" => t["orcid"].downcase
+          }
+        end
+
+        transformed.compact
+
+      rescue JSON::ParserError
```
okeen, line 27: Maybe move this exception catch right after `obj = JSON.parse(json)`? It's the only line that can throw that error and would work like a guard of the method.

okeen (approved):
Looks great!
Just a small improvement suggested, but nothing blocking.

app/services/hyku_addons/bolognese/json_fields_reader.rb
prdanelli: Thanks Eneko, in this situation I wanted to return the json string back if there was an exception, rather than continuing execution of the later statements.

app/services/bolognese/writers/ris_writer_behavior.rb
prdanelli: Thanks @okeen. Thats actually part of the RIS format specificiation, dont ask me why :)

PR #393, opened 2021-07-06
AccountBehavior being dropped from engine

prdanelli:
AccountBehavior seems to be dropped from the engine, so thereby loses the ability to call certain functions on the Account model. 

This is possibly by I am seeing issues with Hyrax DOI update actors being called as randomly the Account will forget that it is required to call those actors. 

I am super unsure about this, but the specs all seem to pass locally.


okeen (approved):
Looks OK, let's see what happens :)

PR #402, opened 2021-07-09
Hotfix/hyrax doi update fix

prdanelli:
Fixes this: https://github.com/samvera-labs/hyrax-doi/pull/33

This should get merged and the Hyrax::DOI::Actor removed when its merged upstream.


okeen (approved):
Looks good mate

PR #403, opened 2021-07-12
Further changes to Anschutz worktype

BertZZ:
Rights Statement bought out of Hyrax/Hyku so that we can use the per tenant authority code


app/models/concerns/hyku_addons/solr_document_behavior.rb
```diff
@@ -124,6 +124,7 @@ def attributes
       attribute :rights_statement_text, SolrDocument::Solr::Array, solr_name('rights_statement_text')
       attribute :qualification_subject_text, SolrDocument::Solr::Array, solr_name('qualification_subject_text')
       attribute :is_format_of, SolrDocument::Solr::Array, solr_name('is_format_of')
+      attribute :is_part_of, SolrDocument::Solr::Array, solr_name('is_format_of')
```
okeen, line 127: Copy Pasta issue here

okeen (changes requested):
I would modify the anschutz work type test to include "is_part_of" too, as that would have detected the C&P issue beflow.

okeen (approved):

PR #422, opened 2021-07-30
Turns version number into a dropdown for 3 Redlands worktypes

BertZZ:


okeen (changes requested):
The code looks ok, but ideally it should have tests checking that the redlands work types are actually reflecting these changes. Remember what happened the day after you left for vacations...

You can add `expect(page).not_to have_selector("#field_id")` overwriting the method `set_work_metadata` at https://github.com/ubiquitypress/hyku_addons/blob/main/spec/features/create_redlands_article_spec.rb

okeen (approved):
Nice, thanks for that!

PR #433, opened 2021-08-16
Dumps facet override into a class_eval in the engine 

BertZZ:


prdanelli (approved):
NOTE: 

This is a quick fix solution, which we want to come back to as keeping code in the engine.rb is not an ideal long term solution.

PR #437, opened 2021-08-18
API fixes (and fixing some spelling issues) 

BertZZ:


app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -111,7 +111,11 @@ json.prerequisites work.try(:prerequisites)
 json.publisher work.publisher
 json.qualification_grantor work.try(:qualification_grantor)
 json.qualification_level work.try(:qualification_level)
-json.qualification_name work.try(:qualification_name)
+
+qualification_name_service = HykuAddons::QualificationNameService.new
```
okeen, line 115: I like writing these as follows:
```suggestion
if (work_id = Array.wrap(work.try(:qualification_name)).first)
  json.qualification_name HykuAddons::QualificationNameService.new.label(id)
end
```
My reasoning is:
1- We leave very clear from the beginning that there is a condition to add the json field ( qualification_name_service )
2- The value of it is now available to use in a variable that you know for certain is NOT nil.
3- It makes it very easy and straightforward to use.

okeen (approved):
Looks good, just a small suggestion for the future

PR #438, opened 2021-08-19
Refactor simplfied_admin_set

prdanelli:
Do not remove the relationships tab fields, just the trigger to view them, otherwise visibility options set in the workflow will not be respected. 

Fixes https://ubiquitypress.atlassian.net/browse/REPO-1348

### DIFF
--- a/app/helpers/hyku_addons/notes_tab_form_helper.rb
+++ b/app/helpers/hyku_addons/notes_tab_form_helper.rb
@@ -3,7 +3,7 @@ module HykuAddons
   module NotesTabFormHelper
     def form_tabs_for(form:)
       if Flipflop.enabled?(:notes_tab_form)
-        super + ['notes']
+        super + ["notes"]
       else
         super
       end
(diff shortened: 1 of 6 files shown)


mankind (approved):
Thanks @prdanelli

app/helpers/hyku_addons/notes_tab_form_helper.rb
```diff
@@ -3,7 +3,7 @@ module HykuAddons
   module NotesTabFormHelper
     def form_tabs_for(form:)
       if Flipflop.enabled?(:notes_tab_form)
-        super + ['notes']
+        super + ["notes"]
```
mankind, line 6: I thought your suggestion was to use double quotes only in contatenation

prdanelli: The other way around, only use single quotes inside of a double quotes string where you are performing a conditional. 

There is no speed different between the two and double quotes allow variables to be parsed, so it means we would only use one two of quote, rather than two types if we used single quotes.

PR #440, opened 2021-08-23
Adds Authentication to the SMTP settings form

BertZZ:


spec/mailers/per_tenant_smtp_mailers_spec.rb
```diff
@@ -59,6 +60,7 @@ def smtp_settings; end
         expect(settings[:user_name]).to eq "username"
         expect(settings[:password]).to eq "password"
         expect(settings[:domain]).to eq "test.custom_domain.com"
+        expect(settings[:authentication]).to eq "login"
```
BertZZ, line 63: I am less sure that this is actually correct now. Might need some more work?

okeen (approved):
Looks good!

PR #286, opened 2021-05-18
REPO-1089: Adds missing cache tags to Hyku Addons'  _tenant.json builder overwrite

okeen:
DO NOT MERGE UNTIL WE RELEASE THE CACHING

PR #278, opened 2021-05-13
REPO:-1089: Adds fragment caching per tenant

okeen:

PR #446, opened 2021-08-31
REPO 1089 Implements Russian Doll Cache to work and Tenant endpoints 

BertZZ:
Co-authored-by: Eneko Taberna <eneko.taberna@gmail.com>
Co-authored-by: Chris Colvard <chris.colvard@gmail.com>

Supersedes #278 and #286


app/lib/hyku_addons/per_tenant_smtp_interceptor.rb
```diff
@@ -10,11 +10,11 @@ def self.delivering_email(message)
       Account.find_by(tenant: Apartment::Tenant.current)&.switch!
       return unless (mailer_settings = Settings.smtp_settings.presence)
 
-      if mailer_settings.from.present?
-        message.from = mailer_settings.from
-        message.reply_to = mailer_settings.from
-        message.return_path = mailer_settings.from
-        message.smtp_envelope_from = mailer_settings.from
+      if (from = mailer_settings.from.presence)
```
prdanelli, line 13: There are a few other places where a similar pattern is employed, but its a little different and IMO a little clearer to read. 

`if (from = mailer_settings.from).present?`

Can we use that style instead so we keep stuff consistent.

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -68,14 +68,29 @@ def reset!
...
+        disable_tenant_cache
       end
 
       def switch_host!(cname)
         Rails.application.routes.default_url_options[:host] = cname
         Hyrax::Engine.routes.default_url_options[:host] = cname
       end
+
+      def setup_tenant_cache
+        return unless Flipflop.enabled?(:cache_enabled).presence
```
prdanelli, line 80: Can you add a space beneath the guard statements to make them stand out from the rest of the business logic? Readability is second only to functionality in importance, so helping users understand what we are doing quickly is important to the project long term. 

Try to use `present?` and `blank?` for guard statments instead of `presence` as that has a different use case where a return value is required, not just a boolean. In this case neither is required as `Flipflop` does that for you.... it should be `return if Flipflop.enabled?(:cache_enabled)`.

So the whole thing would be:

```
      def setup_tenant_cache
        return unless Flipflop.enabled?(:cache_enabled)

        Rails.application.config.action_controller.perform_caching = true
        ActionController::Base.perform_caching = true
        Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id }
        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
      end
```

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -68,14 +68,29 @@ def reset!
...
+        Rails.application.config.action_controller.perform_caching = true
+        ActionController::Base.perform_caching = true
+        Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id }
+        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
+      end
+
+      def disable_tenant_cache
+        Rails.application.config.action_controller.perform_caching = Settings.cache_enabled || false
```
prdanelli, line 88: I missed this the first time i read the PR... the `|| false` at the end. 

What is `cache_enabled` returning unless its a boolean? If its not been set, then it would likely be `nil`, which is `falsey` and so equates to `false` when evaluated. 

If you desperately need the false there, which i cant see why you would, can you make it clearer what its doing by surrounding it with `()`, like: 

`Rails.application.config.action_controller.perform_caching = (Settings.cache_enabled || false)`

Now its super clear that its either of these values, depending on what is set.

config/features.rb
```diff
@@ -28,4 +28,8 @@
   feature :notes_tab_form,
           default: false,
           description: "Enables the Notes tab in the deposit form"
+
+  feature :cache_enabled,
+          default: false,
+          description: "TURBO MODE. Turns on cache for API endpoints. Experimental"
```
prdanelli, line 34: Can you remove the "TURBO MODE.", its not necessary.

app/models/concerns/hyku_addons/account_behavior.rb
prdanelli: Should this be `Flipflop.enabled?(:cache_enabled)` not `Settings.cache_enabled`?

config/features.rb
```diff
@@ -28,4 +28,8 @@
   feature :notes_tab_form,
           default: false,
           description: "Enables the Notes tab in the deposit form"
+
+  feature :cache_enabled,
```
prdanelli, line 32: Can the feature be renamed to something more specific, rather than "cache_enabled". It's a big system and there could be more caches added in the future, so this should be "api_caching"  or similar.

spec/features/create_collection_spec.rb
```diff
@@ -65,6 +66,7 @@
     let(:user) { FactoryBot.create(:user) }
```
prdanelli, line 66: You don't need `FactoryBot` here, the create method is global within the specs.

prdanelli (changes requested):

app/models/concerns/hyku_addons/account_behavior.rb
BertZZ: I worked on this with @okeen but can't remember why we had to keep it like that

config/features.rb
BertZZ: Spoilsport :P

app/models/concerns/hyku_addons/account_behavior.rb
prdanelli: There is nthing else in this PR that matches that, is it elsewhere?

BertZZ: It was set in the docker-compose YAML but I am pretty sure it is resolving to to nil now

BertZZ: Unless I was supposed to bring it over and didn't.............

BertZZ: Yeah we had `SETTINGS__CACHE_ENABLED: "true"` in the docker-compose

prdanelli: Is that going to be in the production YAML too? If so, you can get rid of the `|| false` as we control the default and it doesn't need to be cluttered up by code that will never be run.

app/lib/hyku_addons/per_tenant_smtp_interceptor.rb
```diff
@@ -8,9 +8,9 @@ def self.available_smtp_fields
 
     def self.delivering_email(message)
       Account.find_by(tenant: Apartment::Tenant.current)&.switch!
-      return unless (mailer_settings = Settings.smtp_settings.presence)
+      return unless (mailer_settings = Settings.smtp_settings).present?
```
prdanelli, line 11: Its best not to negate a positive, so this should be `return if ().blank?`

config/features.rb
```diff
@@ -29,7 +29,7 @@
           default: false,
           description: "Enables the Notes tab in the deposit form"
 
-  feature :cache_enabled,
+  feature :api_cache_enabled,
```
prdanelli, line 32: You don't need the `_enabled` suffix as the feature will be enabled or disabled, so you would be disabling the "api_cache_enabled" which hurts my brain. 

Just have it as "api_cache", which is then enabled or disabled by the flipflop.

prdanelli (changes requested):

app/models/concerns/hyku_addons/account_behavior.rb
okeen: The reason why we use `Settings` when resetting the caching config is to set it by default to whatever the default application wide configuration is. This way you can enable the cache in a per tenant way and still have a default configuration to rely on for every tenant.

Also, consider that at that point, there is no active tenant, which means the FlipFlop features are not going to be loadable. We need to rely on Rails' env settings for that.

app/lib/hyku_addons/per_tenant_smtp_interceptor.rb
okeen: We are trying to both do the conditional clause evaluation and assign it to a variable in a single step.

You can't use the `present?` method for that as it returns a truthy or falsey instead of the hash we are interested on, we need to use that hash to do some stuff later.

app/models/concerns/hyku_addons/account_behavior.rb
okeen: Also, the reason why I wanted to ensure it was either `true` or `false` was to not mess up too much with ActionController, but I think you are right and it can safely be a nil.

prdanelli: @okeen if you're using `present?` and `blank?` then `nil` is fine. I just wanted to check that it is actually a variable on the `Settings` object and isn't going to return a "method not found error".

app/lib/hyku_addons/per_tenant_smtp_interceptor.rb
prdanelli: @okeen `.presence` would return the value or `nil`, so the only reason i can see that it would be needed inside the `()` is that you don't know if `smtp_settings` isn't actually set or you need to add an additional stage to the condition `...presence || backup_var`. 

>You can't use the present? method for that as it returns a truthy or falsey instead of the hash we are interested on, we need to use that hash to do some stuff later.

If you're just assigning the value and want to check its not `nil`, then using `.present?` is clearer when reading the code IMO - I use it a lot for single line guard statements that also assign a variable. 

Its the brackets that that work out the order of evaluation... `foo = bar.present?` would be `true`/`false`, but `(foo = bar).present?` assigns `foo` to the value of `bar` and checks if they are present.

docker-compose.yml
```diff
@@ -102,6 +102,7 @@ services:
       SECRET_KEY_BASE: asdf
       SETTINGS__ACTIVE_JOB__QUEUE_ADAPTER: sidekiq
       SETTINGS__BULKRAX__ENABLED: "true"
+      SETTINGS__CACHE_ENABLED: "true"
```
prdanelli, line 105: Can this be removed now?

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -74,8 +75,27 @@ def switch_host!(cname)
...
+
+      def disable_tenant_cache
+        Rails.application.config.action_controller.perform_caching = api_cache?
+        ActionController::Base.perform_caching = api_cache?
+
+        Rails.application.config.cache_store = :file_store, Settings.cache_filesystem_root
+        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
+      end
+
+      def api_cache?
```
prdanelli, line 95: Now that we have renamed the flipflop, we should rename this method to match, so we dont have `api_cache` and `cache_api`, which will be hard to remember which is which. Can this become `cache_api?`.

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -74,8 +75,27 @@ def switch_host!(cname)
...
         Hyrax::Engine.routes.default_url_options[:host] = cname
       end
+
+      def setup_tenant_cache
+        Rails.application.config.action_controller.perform_caching = api_cache?
+        ActionController::Base.perform_caching = api_cache?
+
+        Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id } if api_cache?
+        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
+      end
```
prdanelli, line 85: Now that we have boiled down this method and seen exactly what is necessary, can I suggest that instead of the mutliple calls to `api_cache?`, we do something like: 

      def setup_tenant_cache
        if api_cache?
          Rails.application.config.action_controller.perform_caching = true
          ActionController::Base.perform_caching = true
          Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id } 
        end

        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
      end

This then makes it clear that if the caching is enabled, then we are altering the default settings for the tenant.

prdanelli (changes requested):

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -74,8 +75,23 @@ def switch_host!(cname)
...
+
+      def setup_tenant_cache(is_enabled)
+        Rails.application.config.action_controller.perform_caching = is_enabled
+        ActionController::Base.perform_caching = is_enabled
+        # rubocop:disable Style/ConditionalAssignment
+        if is_enabled
+          Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id }
+        else
+          Rails.application.config.cache_store = :file_store, Settings.cache_filesystem_root
+        end
```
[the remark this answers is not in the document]
BertZZ: Its a dumb cop, but ok

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -59,9 +58,11 @@ def switch!
         datacite_endpoint.switch!
         Settings.switch!(name: locale_name, settings: settings)
         switch_host!(cname)
+        setup_tenant_cache(cache_api?)
       end
 
       def reset!
+        setup_tenant_cache(cache_api?)
```
prdanelli, line 65: This should be `false` no?

BertZZ: No I think I just want to make everything dependent on the Flipflop otherwise you turn off the Cashing after every request

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -74,8 +75,24 @@ def switch_host!(cname)
...
+        if is_enabled
+          Rails.application.config.cache_store = :redis_cache_store, { url: Redis.current.id }
+        else
+          Rails.application.config.cache_store = :file_store, Settings.cache_filesystem_root
+        end
+        # rubocop:enable Style/ConditionalAssignment
+        Rails.cache = ActiveSupport::Cache.lookup_store(Rails.application.config.cache_store)
+      end
+
+      def cache_api?
```
prdanelli, line 92: If we are only using this method one it might be worth removing it and using the Flipflop directly.

prdanelli (approved):

app/models/concerns/hyku_addons/account_behavior.rb
BertZZ: Might be useful in the future? Not sure it's big enough to worry about now

PR #448, opened 2021-09-01
Adds method to SolrService that automatically handles pagination of results

okeen:


okeen:
Example code that builds a CSV file with the returning set:

require 'csv'

CSV.open("tmp/#{account_name}_file_sizes.csv", "w+") do |csv_file|
  csv_file << %w[file_id file_size]
  ActiveFedora::SolrService.find_each 'has_model_ssim:FileSet', rows: 100 do |d|
    csv_file << [d["id"], d["file_size_lts"]]
  end
end

okeen:
There is already a better implementation of this at ActiveFedora::find_each or SolrService#search_in_batches

PR #450, opened 2021-09-07
Failing DOI Minting spec

prdanelli:
X-out the failing spec as we can't seem to get it working via the feature specs and the account class variables that are used to integrate the gem


okeen (approved):
Looks OK mate!

PR #451, opened 2021-09-07
REPO-1361: Adds rake task and service object to disable every tenant cache

okeen:


app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -79,9 +79,9 @@ def switch_host!(cname)
       def setup_tenant_cache(is_enabled)
         Rails.application.config.action_controller.perform_caching = is_enabled
         ActionController::Base.perform_caching = is_enabled
-        # rubocop:disable Style/ConditionalAssignment
```
BertZZ, line 82: :pray:

app/services/hyku_addons/cache_expiration_service.rb
```diff
@@ -0,0 +1,20 @@
...
+        Rails.logger.info "Expiring cache for #{account.cname}"
+        next if account.redis_endpoint.is_a?(NilRedisEndpoint)
+
+        expire_cache_for(account)
+      end
+    end
+
+    def expire_cache_for(account)
+      account.redis_endpoint.switch!
+      account.setup_tenant_cache(true)
```
BertZZ, line 16: Could we not call account.switch! and save a line?

spec/models/concerns/hyku_addons/account_behavior_spec.rb
```diff
@@ -43,7 +44,7 @@
       end
     end
 
-    context "when cashe is disabled" do
+    context "when cache is disabled" do
```
BertZZ, line 47: :scream:

spec/services/hyku_addons/cache_expiration_service_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+  end
+
+  describe "perform" do
+    before do
+      allow(service).to receive(:expire_cache_for)
+      service.perform
+    end
+
+    it "updates the frontend_url of the Account" do
+      expect(service).to have_received(:expire_cache_for).once
```
BertZZ, line 23: What does this have to do with the frontend URL?

app/services/hyku_addons/cache_expiration_service.rb
okeen: There are some extra things made in `account.switch!` that we really don't need here, including loading the settings which can be costly.
This should avoid that

spec/services/hyku_addons/cache_expiration_service_spec.rb
okeen: Copy pasta thing, good catch

BertZZ (approved):

PR #452, opened 2021-09-08
Move Engine class_eval's to behaviors

BertZZ:
The HykuAddons Engine had become a dumping ground for class_eval monkey patches and initializers which were hard to read, understand and edit. This PR moves as much as possible from the engine.rb to behaviors and initializer files that are self contained and mean that the engine file went from 700 LOC to ~200. 

One consequence of the changes is that references to the `Bulkrax` module now need to be from the root namespace, so all references to the module now need to be `::Bulkrax`.


app/parsers/hyku_addons/csv_parser.rb
```diff
@@ -45,17 +45,17 @@ def retrieve_cloud_files(files)
...
         # Problematic for a large upload
-        Bulkrax::DownloadCloudFileJob.perform_now(file, target_file)
+        ::Bulkrax::DownloadCloudFileJob.perform_now(file, target_file)
       end
       nil
     end
 
     def entry_class
-      HykuAddons::CsvEntry
+      CsvEntry
```
thenapking, line 54: I know this strictly doesn't need `HykuAddons::` prepended, but since there is a Bulkrax class of exactly the same name it really helps the reader understand which one is being referred to.  Same with the `admin_set_entry_class`.

app/models/concerns/hyku_addons/mailboxer_message_mailer_behavior.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  module MailboxerMessageMailerBehavior
+    extend ActiveSupport::Concern
+
+    def new_message_email(message, receiver)
```
thenapking, line 7: I know Codeclimate wants us to combine these into one method, but that seems extreme.  If you can get CodeClimate to ignore this I'd be happy with that.

app/models/concerns/hyku_addons/bulkrax/object_factory_behavior.rb
```diff
@@ -0,0 +1,112 @@
...
+
+      # Override if we need to map the attributes from the parser in
+      # a way that is compatible with how the factory needs them.
+      def transform_attributes
+        if klass == ::Collection || klass == ::AdminSet
+          attributes.slice(*permitted_attributes)
+        else
+          attrs = attributes.slice(*permitted_attributes).merge(file_attributes)
+          attrs = attrs.merge("file_set" => attributes["file_set"])
+          attrs["uploaded_files"].each_with_index { |id, i| attrs["file_set"][i]["uploaded_file_id"] = id if attrs["file_set"][i].present? } if attrs["file_set"].present?
```
thenapking, line 106: We talked about this line on Slack, and I proposed a refactor.  I think something along the lines of this almost gets there, but it might need a bit more work:
```
file_count = attrs['uploaded_files'].size
attrs["file_set"] = Hash[(0..file_count).zip attrs['uploaded_files']]
````

File: app/models/concerns/hyku_addons/bulkrax/object_factory_behavior.rb [hunk 4]
```diff
@@ -0,0 +1,112 @@
...
+
+      # rubocop:disable Metrics/PerceivedComplexity
+      # rubocop:disable Metrics/CyclomaticComplexity
+      def update
+        raise "Object doesn't exist" unless object
+
+        destroy_existing_files if @replace_files && (klass != ::Collection || klass != ::AdminSet)
+        attrs = update_attributes
+        run_callbacks :save do
+          if klass == ::AdminSet
```
thenapking, line 53: This bit is common to both update and create so could be moved into another method to reduce the complexity.

spec/services/hyku_addons/active_support_inflectors_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+
+require "rails_helper"
+
+RSpec.describe ActiveSupport::Inflector do
+  describe "media" do
+    it "pluralizes" do
+      expect("media".pluralize).to eq "medias"
+    end
+
+    it "singgularizes" do
```
thenapking, line 11: Spelling mistaked.  To many `g`s

spec/services/bulkrax/bulkrax_configuration_spec.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+# This spec is to ensure that configuration removed from the Engine is still being read as it was before the change
+RSpec.describe Bulkrax do
```
thenapking, line 4: Seems a bit unnecessary to test what is essentially a configuration.  Can't the Bulkrax config go in an initializer not a service?

spec/features/bulkrax_import_spec.rb
```diff
@@ -274,7 +270,7 @@
           let(:collection_2) { create :collection, id: "bedd7330-5040-4687-8226-0851f7256dff", title: ["This Title Should Change"] }
 
           it "updates the collections" do
-            perform_enqueued_jobs(only: [Bulkrax::ImporterJob, HykuAddons::ImportWorkCollectionJob]) do
+            perform_enqueued_jobs(only: [Bulkrax::ImporterJob, Bulkrax::ImportWorkCollectionJob]) do
```
thenapking, line 273: `Bulkrax::ImportWorkCollectionJob` shouldn't be called here, the HykuAddons one should.  Our one sets collection titles, and the Bulkrax ones doesn't.  If the HykuAddons one isn't being called any more we have a problem.

config/initializers/quick_classification_bug_fix.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+# NOTE: This issue only seems to present in development, and not consistently.
+# Compact the available work types to remove `nil` and prevent an `Nil location provided. Can't build URI.` error
+# when not all options are selected, thrown from `SelectTypePresenter#switch_to_new_work_path`
+Hyrax::QuickClassificationQuery.class_eval do
+  def normalized_model_names
+    models.map { |name| concern_name_normalizer.call(name) if Site.first.available_works.include?(name) }.compact
```
thenapking, line 7: Do we know if this bug fix is still required, and do we have tests which relate to it?

app/parsers/hyku_addons/csv_parser.rb
prdanelli: I do agree that understanding of the code is paramount, but if its namespaced `HykuAddons` within a `HykuAddons` module, then it seems like a smell. What do you think? Is it better to present "proper" code and reduces as much redundancy as possible, or to explain as much as is possible to the reader?

app/models/concerns/hyku_addons/user_behavior.rb
```diff
@@ -14,6 +14,10 @@ module UserBehavior
       scope :with_public_profile, -> { where(display_profile: true) }
     end
 
+    def mailboxer_email(_obj)
```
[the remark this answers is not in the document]
prdanelli: It's not by the look of it. I'll remove it.

app/models/concerns/hyku_addons/mailboxer_message_mailer_behavior.rb
prdanelli: Refactored 👍🏻

app/actors/hyku_addons/actors/file_set_actor_behavior.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  module Actors
+    module FileSetActorBehavior
+      def attach_to_work(work, file_set_params = {})
+        acquire_lock_for(work.id) do
```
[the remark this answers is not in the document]
prdanelli: AFAIK, this was added so we could add another callback `Hyrax.config.callback.run(:task_master_after_create_fileset, file_set, user)` to the after fileset actor hook.

app/models/concerns/hyku_addons/bulkrax/object_factory_behavior.rb
prdanelli: As I dont understand the code fully, or if there are good enough tests around this section to properly refactor it, i went with the simple solution. 

```          attrs["uploaded_files"].each_with_index do |id, i|
            attrs["file_set"][i]["uploaded_file_id"] = id if attrs["file_set"][i].present?
          end if attrs["file_set"].present?
```

thenapking (approved):
LGTM

PR #453, opened 2021-09-08
Hyrax Orcid Gem Integration

prdanelli:
Integrate the Hyrax Orcid Gem into Hyku Addons


lib/hyku_addons/engine.rb
```diff
@@ -513,6 +519,9 @@ def self.dynamically_include_mixins
       # Insert at the end of the actor chain
       Hyrax::CurationConcern.actor_factory.use HykuAddons::Actors::TaskMaster::WorkActor
 
+      # Remove the Hyrax Orcid JSON Actor as we have our own
+      Hyrax::CurationConcern.actor_factory.middlewares.delete(::Hyrax::Actors::Orcid::JSONFieldsActor)
```
mankind, line 523: 1. I can't see **Hyrax::Actors::Orcid::JSONFieldsActor** in hyrax or hyku_addons. Instead i saw one with a different name here:  **HykuAddons::Actors::Orcid::JSONFieldsActor**
https://github.com/ubiquitypress/hyku_addons/blob/main/app/actors/hyku_addons/actors/json_fields_actor.rb

2. Do we really need to delete it if the version in hyrax-orcid is direct copy of the one in hyku_addons.
If we need to adjust that actor for hyku_addons do we need to always refer to hyrax-orcid

app/presenters/concerns/hyku_addons/person_or_organization_presenter_behavior.rb
```diff
@@ -81,12 +83,13 @@ def wikidata_link
...
-        return [] unless send(field)&.first.present?
+        return if (field_value = solr_document.public_send(field).first.presence || "[]").blank?
 
-        JSON.parse(send(field).first).collect do |hash|
+        JSON.parse(field_value).collect do |hash|
           name = hash.slice("#{field}_family_name", "#{field}_given_name", "#{field}_organization_name").values.map(&:presence).compact.join(', ')
           PersonOrOrganization.new(name, hash["#{field}_orcid"], hash["#{field}_isni"], hash["#{field}_ror"], hash["#{field}_grid"], hash["#{field}_wikidata"])
         end
       end
+      alias participants person_or_organization_list
```
mankind, line 93: **person_or_organization_list** defined elsewhere or it is being defined here for the first time. If so why rename it with alias, while not call it particpants

prdanelli: the method `participants` is defined inside of Hyrax Orcid, but some essentialy the same thing as `person_or_organization_list`, so i just aliased it to the existing method here.

app/services/bolognese/writers/orcid/hyku_addons_xml_builder.rb
```diff
@@ -0,0 +1,100 @@
...
+
+            xml_subtitle
+          end
+
+          xml_short_description
+          xml_work_type
+          xml_date_published
+
+          # NOTE: A full list of external-id-type: https://pub.orcid.org/v2.1/identifiers
+          @xml[:common].send("external-ids") do
```
mankind, line 21: Do you want to consider extracting this and the block underneath to standalone methods and call those here. 
It follows the pattern of the methods calls above where you called things like xml_date_published.

Additionally, it meant eliminate the need for rubocop override **rubocop:disable Metrics/MethodLength**

okeen (dismissed):
This is great work, mate.

app/services/bolognese/writers/orcid/hyku_addons_xml_builder.rb
prdanelli: As much as I hate changing structure to appease Rubocop, I agree that isn't bad enough for me not to do it :)

app/helpers/hyku_addons/orcid_helper_behavior.rb
```diff
@@ -0,0 +1,23 @@
...
+  module OrcidHelperBehavior
+    extend ActiveSupport::Concern
+
+    include Hyrax::Orcid::HelperBehavior
+
+    # Override the Hyrax::Orcid JSON Fields creator/contributor helper, method which uses different participant name keys
+    def participant_to_string(type, participants)
+      return "-" if participants.blank?
+
+      participants
```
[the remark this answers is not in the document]
prdanelli: I think the guard clause being here means i don't need to add a fallback within the lower block.

PR #455, opened 2021-09-10
REPO-1251 Fixes account creation and cleanup jobs.

okeen:


app/jobs/concerns/hyku_addons/cleanup_account_job_behavior.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+module HykuAddons
+  module CleanupAccountJobBehavior
+    extend ActiveSupport::Concern
+
+    def perform(account)
+      cleanup_solr(account)
+      cleanup_fedora(account)
+      cleanup_redis(account)
+      tenant = account.tenant
```
prdanelli, line 10: Can you add a comment here outlining what is going on?

app/jobs/concerns/hyku_addons/cleanup_account_job_behavior.rb
```diff
@@ -0,0 +1,23 @@
...
+      account.destroy
+      cleanup_database(tenant)
+    end
+
+    private
+
+      def cleanup_database(tenant)
+        Apartment::Tenant.drop(tenant)
+      rescue StandardError
+        nil # ignore if account.tenant missing
```
prdanelli, line 20: This `nil` isn't necessary as returning nothing is `nil`. 

```
      def cleanup_database(tenant)
        # ignore if account.tenant missing
        Apartment::Tenant.drop(tenant) rescue StandardError
      end
```

This would be a nicer way of writing it - inlining the `rescue`. 

It's also a really bad idea to catch the low level errors, is there not a more specific error, or set of multiple error we can catch instead?

prdanelli: Can you test thing to make sure it works ok, I'm pretty sure it should :)

prdanelli (changes requested):

app/jobs/concerns/hyku_addons/cleanup_account_job_behavior.rb
okeen: Well, I left it as it original was in Hyku to avoid making more changes that we strictly needed, but I think your proposal should be logically the same. The tests will tell :)

okeen: Rubocop doesnt like not handling the rescues: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/RescueModifier

prdanelli (approved):

PR #468, opened 2021-09-16
cross tenant search implementation

mankind:


app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -106,6 +109,25 @@ def public_settings
...
+    end
+
+    def add_parent_id_to_child
+      create_child_parent_association_from_submitted_tenant_ids
+      settings['tenant_list'] = []
+      save
+    end
+
+    def remove_existing_child_records
+      return self unless children.present?
```
okeen, line 127: I think you can change the method for the following:

```suggestion
      self.children = []
      #Not sure if this is necessary
      save
```

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -140,5 +164,21 @@ def set_smtp_settings
           HykuAddons::PerTenantSmtpInterceptor.available_smtp_fields.each_with_object("").to_h
         )
       end
+
+      def set_shared_search_default
+        return if settings['shared_search'].present?
+        self.shared_search = false
```
okeen, line 170: I think you can safely change both lines for:
```suggestion
        self.shared_search ||= false
```

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -140,5 +164,21 @@ def set_smtp_settings
...
+      end
+
+      def set_default_tenant_list
+        return if settings['tenant_list'].present?
```
okeen, line 174: Same as before

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -140,5 +164,21 @@ def set_smtp_settings
...
+        self.tenant_list = []
+      end
+
+      def create_child_parent_association_from_submitted_tenant_ids
+        tenant_list.each do |tenant|
```
okeen, line 179: Let's use Rails' scopes to help us again:
```suggestion
        self.children = Account.where(name: tenant_list).all
```

app/models/concerns/hyku_addons/solr_endpoint_override.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+# Remove the solr collection then destroy this record
+module HykuAddons
+  module SolrEndpointOverride
+    extend ActiveSupport::Concern
+
+    def remove!
+      # Spin off as a job, so that it can fail and be retried separately from the other logic.
+      if account.shared_search_enabled?
```
okeen, line 10: Is this maybe clearer?

```suggestion
      RemoveSolrCollectionJob.perform_later(collection, connection_options, account.shared_search_enabled?)
```

app/views/_controls.html.erb
```diff
@@ -0,0 +1,26 @@
+<!-- copied from https://raw.githubusercontent.com/samvera/hyrax/main/app/views/_controls.html.erb -->
+
+<nav class="navbar navbar-default navbar-static-top" role="navigation" aria-label="Root Menu">
+  <div class="container-fluid">
+    <div class="row">
+      <ul class="nav navbar-nav col-sm-5">
+        <li <%= 'class=active' if current_page?(hyrax.root_path) %>>
```
okeen, line 7: There is a lot of repeated code here. Do you think we can make a helper method that does all this `li` and `link_to` configuration? Let me know if you  want to pair!

app/views/proprietor/accounts/_add_data_for_cross_tenant_search.html.erb
```diff
@@ -0,0 +1,25 @@
+
+
+<p>
+  <%= f.check_box :shared_search, {checked: account.shared_search_tenant?, name: "account[settings][shared_search]", id: "account_settings_shared_search"}, true , false %>
```
okeen, line 4: Is Rubocop happy with this line?
```suggestion
  <%= f.check_box :shared_search, { checked: account.shared_search_tenant?, name: "account[settings][shared_search]", id: "account_settings_shared_search"}, true , false %>
```

Also, I personally like creating helper methods for these, even if they are for one time use. Leaves the views much cleaner, which is an issue with ERB:
```
<%= enable_shared_search_check_box(f) %>
```
and then in the helper:

```
def enable_shared_search_check_box(f, account)
    f.check_box :shared_search, {checked: account.shared_search_tenant?, name: "account[settings][shared_search]", id: "account_settings_shared_search"}, true , false
end
```

app/views/proprietor/accounts/_add_data_for_cross_tenant_search.html.erb
```diff
@@ -0,0 +1,25 @@
...
+  <%= f.label "Search Only(please tick checkbox select tenants)" %>
+</p>
+<br />
+
+<%= f.label :tenant_list %>
+
+<p>
+  <%  @account.children.each do |tenant| %>
+    <%= check_box_tag "account[settings][tenant_list][]", tenant.tenant, @account.children.include?(tenant) %>
```
okeen, line 13: This checkbox tag could go in a helper as well

app/views/proprietor/accounts/_shared_search_toggle.html.erb
```diff
@@ -0,0 +1,71 @@
...
+                <th class="description"><%= t('.description') %></th>
+                <th class="action"><%= t('.action') %></th>
+              </tr>
+            </thead>
+            <tbody>
+            <%# grouped_features&.each do |group, features| -%>
+            <% grouped_features&.each do |feature| -%>
+
+              <% if feature_set.grouped? -%>
+                <tr class="group">
```
okeen, line 25: I would extract the `tr` into a partial to make the file smaller and easier to read

app/views/proprietor/accounts/_shared_search_toggle.html.erb
```diff
@@ -0,0 +1,71 @@
...
+                  <td></td>
+                  <td class="name" colspan="<%= 2 + feature_set.strategies.size -%>">
+                    <h2>
+                      <%# t(group ? group.name : :default, scope: [:flipflop, :groups], default: group ? group.title : nil) -%>
+                    </h2>
+                  </td>
+                </tr>
+              <% end -%>
+              <%# features&.each do |feature| %>
+              <tr data-feature="<%= feature.name.dasherize.parameterize %>">
```
okeen, line 35: This could also go into a partial to make the parent one easier to read

app/views/proprietor/accounts/_shared_search_toggle.html.erb
```diff
@@ -0,0 +1,71 @@
...
+                <td class="status">
+                  <span class="<%= feature_set.status(feature) -%>"><%= feature_set.status(feature) -%></span>
+                </td>
+                <td class="name"><%= feature.name.humanize -%></td>
+                <td class="description"><%= feature.description -%></td>
+
+                <% feature_set.strategies.each do |strategy| -%>
+                  <% next unless strategy.is_a? Flipflop::Strategies::ActiveRecordStrategy %>
```
okeen, line 43: maybe this line and its top one should go into a helper method like:
```
def per_valid_feature_set_strategy(feature_set)
    feature_set.strategies.each do |strategy|
       yield(strategy) if strategy.is_a? Flipflop::Strategies::ActiveRecordStrategy
    end 
end
```

okeen (commented):
You did a really nice feature here touching the just the bits you needed to! I've left a few comments here and there but detected nothing to block this getting merged.

app/controllers/concerns/hyku_addons/proprietor_controller_override.rb
```diff
@@ -0,0 +1,36 @@
...
+module HykuAddons
+  module ProprietorControllerOverride
+    extend ActiveSupport::Concern
+
+    def update
+      respond_to do |format|
+        if @account.update(account_params)
+
+          # handles updating of shared search tenant
+          CreateSolrCollectionJob.perform_now(@account)
```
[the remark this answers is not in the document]
mankind: I can add a check there since it will make it more obvious but to directly answer you, CreateSolrCollectionJobOverride ensures no action is taken when it is normal  tenant because it check if a solr  collection exists for that account and does nothing if a solr collection exists and it is a normal tenant

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+module HykuAddons
+  module CrossTenantSharedSearchHelper
+    def redirect_to_shared_search_page
+      cname = current_account.parent&.cname
+      return unless cname.present?
+
+      local_url = "#{request.protocol}#{cname}:#{request.port}/catalog?utf8=✓&locale=en&search_field=all_fields&q="
+      production_url = "#{request.protocol}#{cname}/catalog?utf8=✓&locale=en&search_field=all_fields&q="
+
+      request.host.include?('hyku.docker') ? local_url : production_url
```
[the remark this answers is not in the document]
mankind: Good suggestion, I will make that change.

app/models/concerns/hyku_addons/account_behavior.rb
```diff
@@ -14,15 +14,18 @@ module AccountBehavior
     PRIVATE_SETTINGS = %w[smtp_settings].freeze
 
     included do
+      scope :not_cross_search_tenants_new_list, -> { where.not('settings @> ?', { shared_search: true }.to_json) }
+      scope :not_cross_search_tenants_edit_list, ->(id) { where.not('settings @> ?', { shared_search: true }.to_json).where.not(id: id) }
```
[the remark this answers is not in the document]
mankind: Ok thanks noted, i will change the name and re-use the scope

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
```diff
@@ -0,0 +1,38 @@
...
+    end
+
+    def tenants_not_in_search(account)
+      @list_for_unsaved_record = Account.not_cross_search_tenants_new_list - account.children
+      @list_without_record_under_edit = Account.not_cross_search_tenants_edit_list(account.id) - account.children
+      account.new_record? ? @list_for_unsaved_record : @list_without_record_under_edit
```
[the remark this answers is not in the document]
mankind: I am happy to move it to the model

app/jobs/concerns/hyku_addons/create_solr_collection_job_override.rb
```diff
@@ -0,0 +1,68 @@
...
+
+# Add ability to create  a real slr collection or a collection alias for a shared search tenant
+module HykuAddons
+  module CreateSolrCollectionJobOverride
+    extend ActiveSupport::Concern
+
+    def perform(account)
+      name = account.tenant.parameterize
+
+      perform_for_cross_search_tenant(account, name) if account.shared_search_tenant? && account.tenant_list.present?
```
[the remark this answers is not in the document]
mankind: If rubocop does not reject it then yes , if it does then we will leave the guard clauses

app/views/proprietor/accounts/_add_data_for_cross_tenant_search.html.erb
mankind: Yes Ruboocp was happy hence no rubocop failures. I will move it to a helper.

app/views/proprietor/accounts/_add_data_for_cross_tenant_search.html.erb
mankind: Ok noted

app/views/proprietor/accounts/_shared_search_toggle.html.erb
mankind: I think we should let this one stay as it is , because I copied this file from hyrax and I want it to stay close to the original  hyrax file,  so in future, if they make a change it will be easier to see the difference and update this file.

app/views/proprietor/accounts/_shared_search_toggle.html.erb
mankind: let's let stay as it is in hyrax, so it is easy to follow changes.

app/views/proprietor/accounts/_shared_search_toggle.html.erb
mankind: let's keep the file as it came from hyrax rather than improve it.

app/models/concerns/hyku_addons/account_behavior.rb
mankind: Sorry, what does this give us?

app/views/_controls.html.erb
mankind: This is another file copied from hyrax which i think should be left as close to the original as possible for easy comparing of both files in future if things deviate.

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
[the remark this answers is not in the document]
mankind: Just a quick update, I have moved the code above to the model,.  tried applying the code  changes suggested above but it does not work the old method.

1  The assumption in your code changes above assumes the scope expects many ids instead of one id.  

2. Expecting many ids does take into account the scope chaining that fetches some of the ids we are now passing as part of the items_to_exclude.

I agree that the method could be simply and will be happy to take a look within the week hopefully.

okeen (approved):
Looks good mate!

PR #476, opened 2021-09-27
Pull in many to many association for cross-search from Hyku

mankind:
Pulling in Rob's Many to Many changes for shared-search from Hyku


db/migrate/20200929084240_add_parent_id_and_data_settings_column_to_account.rb
```diff
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 class AddParentIdAndDataSettingsColumnToAccount < ActiveRecord::Migration[5.2]
   def change
-    add_reference :accounts, :parent, foreign_key: { to_table: :accounts }
+    # add_reference :accounts, :parent, foreign_key: { to_table: :accounts }
```
prdanelli, line 4: This comment can be removed.

spec/models/account_cross_search_spec.rb
```diff
@@ -0,0 +1,12 @@
+# frozen_string_literal: true
+require 'rails_helper'
+
+RSpec.describe AccountCrossSearch, type: :model do
+  it do
+    is_expected.to belong_to(:full_account)
+      .class_name('Account')
+
+    is_expected.to belong_to(:search_account)
+      .class_name('Account')
```
prdanelli, line 10: Can these be on a single line to make it clearer what they are doing?

```
is_expected.to belong_to(:search_account).class_name('Account')
```

spec/requests/catalog_controller_spec.rb
```diff
@@ -49,7 +62,8 @@
 
     context 'can fetch data from other tenants' do
       xit 'cross-search-tenant can fetch all record in child tenants' do
-        allow(blacklight_solr_repository).to receive(:build_connection).and_return(RSolr.connect(url: 'http://solr:8983/solr/hydra-cross-search-tenant'))
+        connection = RSolr.connect(url: 'http://solr:8983/solr/hydra-cross-search-tenant')
```
prdanelli, line 65: Might be best to put the setup for this in a `let`.

db/migrate/20200929084240_add_parent_id_and_data_settings_column_to_account.rb
mankind: Thanks done.

spec/models/account_cross_search_spec.rb
mankind: Thanks done.

spec/requests/catalog_controller_spec.rb
mankind: Since this test is **xit**, it is better to leave it till i revisit the whole test.  Thanks.

PR #477, opened 2021-09-27
Suppress deprecation warnings.

prdanelli:
Try and quieten down Ruby and Rails until we need to address these warnings

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -81,7 +81,7 @@ services:
       dockerfile: Dockerfile
       target: dev
     image: hyku_addons/dev:latest
-    command: bash -c "rm -f spec/internal_test_hyku/tmp/pids/server.pid && bundle exec rails server -b 0.0.0.0"
+    command: bash -c "rm -f spec/internal_test_hyku/tmp/pids/server.pid && bundle install && bundle exec rails server -b 0.0.0.0"
     #env_file:
     #  - ./spec/internal_test_hyku/.env
     environment: &web_environment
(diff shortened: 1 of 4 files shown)


docker-compose.yml
```diff
@@ -81,7 +81,7 @@ services:
       dockerfile: Dockerfile
       target: dev
     image: hyku_addons/dev:latest
-    command: bash -c "rm -f spec/internal_test_hyku/tmp/pids/server.pid && bundle exec rails server -b 0.0.0.0"
+    command: bash -c "rm -f spec/internal_test_hyku/tmp/pids/server.pid && bundle install && bundle exec rails server -b 0.0.0.0"
```
prdanelli, line 84: By having this here (and in the sidekiq container as well), we do not have to build the container when a Gem is added/removed/updated. The actual `bundle install` takes a few seconds when there is nothing to change, so it doesn't add much to the application boot time.

docker-compose.yml
```diff
@@ -119,6 +119,7 @@ services:
       GCLOUD_ZONE: 1233465
       GCLOUD_PROJECT_ID: ABCDEFG
       PUBSUB_SERVICEACCOUNT_KEY:
+      RUBYOPT: "-W:no-deprecated -W:no-experimental"
```
prdanelli, line 122: Tell Ruby not to display depreciation warnings.

spec/rails_helper.rb
```diff
@@ -20,6 +20,9 @@
 # Optional execution of specs for examples that fail randomly on CI
 require File.expand_path('support/optional_example', __dir__)
 
+# Try and suppress depreciation warnings
+ActiveSupport::Deprecation.silenced = true
```
prdanelli, line 24: Suppress deprecation warnings in Test logs.

okeen (approved):

PR #494, opened 2021-10-04
Improves worktype specs to check fields are rendered 

BertZZ:


spec/features/create_anschutz_work_spec.rb
```diff
@@ -1,20 +1,102 @@
...
+                                  qualification_name is_format_of source_identifier]
+      worktype_simple_fields.each do |field|
+        expect(page).to have_field("#{work_type}_#{field}")
+      end
+    end
+
+    it "renders all complex fields" do
+      worktype_complex_fields = %w[creator]
+      worktype_complex_fields.each do |field|
+        expect(page).to have_field("#{work_type}_#{field}__#{field}_family_name")
```
prdanelli, line 88: We're creating the field name manually a lot here, can we replace the `"#{work_type}_#{field}__#{field}"` with a `let(:field_name)` at the top?

This example could then be: 

```expect(page).to have_field("#{field_name}_family_name")```

File: spec/features/create_anschutz_work_spec.rb [hunk 2]
```diff
@@ -1,20 +1,102 @@
...
+    click_link "Files" # switch tab
+    expect(page).to have_content "Add files"
+    expect(page).to have_content "Add folder"
+    within('span#addfiles') do
+      attach_file("files[]", Rails.root.join('spec', 'fixtures', 'hyrax', 'image.jp2'), visible: false)
+      attach_file("files[]", Rails.root.join('spec', 'fixtures', 'hyrax', 'jp2_fits.xml'), visible: false)
+    end
+  end
+
+  it 'applys work visibility' do
```
prdanelli, line 43: It doesn't seem to make sense here to use a `case` statement for a single value. Instead, we could use: 

```
  it 'applys work visibility' do
    visit "concern/#{work_type.to_s.pluralize}/new"

    visibility = :open
    find('body').click
    choose("#{work_type}_visibility_#{visibility}")

    expect(page).to have_content('Please note, making something visible to the world (i.e. marking this as Public) may be viewed as publishing which could impact your ability to')
  end
```

spec/features/create_anschutz_work_spec.rb
```diff
@@ -1,20 +1,102 @@
...
+      agent_id: user.user_key,
+      access: 'deposit'
+    )
+    login_as user
+  end
+
+  it 'renders the new Anschutz Work page' do
+    visit "concern/#{work_type.to_s.pluralize}/new"
+    expect(page).to have_content "Add New Anschutz Work"
+  end
```
prdanelli, line 30: Can we increase the readability of the specs by seperating the setup and the expectations?

```
  it 'renders the new Anschutz Work page' do
    visit "concern/#{work_type.to_s.pluralize}/new"

    expect(page).to have_content "Add New Anschutz Work"
  end
```

spec/features/create_anschutz_work_spec.rb
```diff
@@ -1,20 +1,102 @@
...
+    find('body').click
+    choose("#{work_type}_visibility_#{visibility}")
+    case visibility
+    when :open
+      expect(page).to have_content('Please note, making something visible to the world (i.e. marking this as Public) may be viewed as publishing which could impact your ability to')
+    end
+  end
+
+  it 'saves the work' do
+    visit "concern/#{work_type.to_s.pluralize}/new"
```
prdanelli, line 55: We are using the same string URL for all of the specs, so instead, lets make it a `let(:new_work_path)` and then we can just have `visit new_work_path` and make it easier to parse.

spec/features/create_denver_book_chapter_spec.rb
```diff
@@ -1,20 +1,110 @@
...
+    it "renders all simple worktype fields" do
+      worktype_simple_fields = %w[title alt_title resource_type institution abstract keyword
+                                  subject_text org_unit book_title alt_book_title edition
+                                  pagination library_of_congress_classification
+                                  isbn publisher place_of_publication license
+                                  rights_holder rights_statement medium language
+                                  time refereed add_info]
+
+      worktype_simple_fields.each do |field|
+        expect(page).to have_field("#{work_type}_#{field}")
```
prdanelli, line 82: Love this.

spec/features/create_denver_book_spec.rb
```diff
@@ -1,20 +1,107 @@
...
+    check('agreement')
+    click_on('Save')
+    expect(page).to have_content('My Test Work')
+    expect(page).to have_content('Public')
+    expect(page).to have_content("Your files are being processed by Hyku in the background.")
+  end
+
+  context "when rendering the form" do
+    before do
+      visit "concern/#{work_type.to_s.pluralize}/new"
```
prdanelli, line 69: We are using a `before` block for this visit, but not for the other `it` blocks above. If we move the `before to after `login_as user` (adding a line break for readibility of course), then we can remove the duplicate statement.

spec/features/create_denver_presentation_material_spec.rb
```diff
@@ -1,20 +1,106 @@
 # frozen_string_literal: true
 require 'rails_helper'
-require File.expand_path('../helpers/create_denver_work_user_context', __dir__)
-
 include Warden::Test::Helpers
 
 # NOTE: If you generated more than one work, you have to set "js: true"
 RSpec.feature 'Create a DenverPresentationMaterial', js: false do
```
prdanelli, line 6: Can you check if `js: false` is set by default? I think it might be, so we could remove these arguments.

prdanelli (changes requested):
This is great. There are some changes I think we could make to make them easier to read and parse, cleaning up some older habbits at the same time.

spec/features/create_anschutz_work_spec.rb
BertZZ: I don't think so because field is undefined outside the loop

prdanelli: For the loops you are probably right, but for the others we could.

spec/features/create_denver_book_chapter_spec.rb
```diff
@@ -41,18 +41,14 @@
   end
 
   it 'applys work visibility' do
-    visit "concern/#{work_type.to_s.pluralize}/new"
     visibility = :open
```
prdanelli, line 44: Don't forget to remove these now as they are redundant.

prdanelli (changes requested):
Just need to get rid of the, now redunant, `visibility` variables.

spec/features/create_denver_book_chapter_spec.rb
BertZZ: Change this line to choose("#{work_type}_visibility_open}") ??

prdanelli: `choose("#{work_type}_visibility_#{visibility}")` could just be `choose("#{work_type}_visibility_open")` and we can get rid of that variable.

BertZZ:
All specs Green. Good to merge? @prdanelli

PR #498, opened 2021-10-05
Passenger Docker Container

prdanelli:
Try and move to use passenger docker container for development and production

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,24 +1,18 @@
-# Ripped off from Avalon
-# Base stage for building gems
-FROM        ruby:2.7.2-buster as bundle
-RUN         echo "deb http://deb.debian.org/debian buster-backports main" >> /etc/apt/sources.list \
-         && apt-get update \
-         && rm -rf /var/lib/apt/lists/* \
-         && apt-get clean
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,100 +1,53 @@
...
-            unzip \
-            && \
-            apt-get clean && \
-            rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
-
-# If changes are made to fits version or location,
-# amend `LD_LIBRARY_PATH` in docker-compose.yml accordingly.
+FROM phusion/passenger-ruby27:2.0.0 as BASE_IMAGE
+
+RUN install_clean --allow-unauthenticated \
```
mankind, line 4: So the missing packages from the old dockerfile have been pre-installed ?

      openssh-client ,  tzdata, nodejs, curl gnupg2 ffmpeg

prdanelli: `gnupg2 ffmpeg` are there, just lower down the list. The others are preinstalled. I also added Vim because its annoying having to install it to debug something inside a Gem. 

The installed packages can be listed with `dpkg -l` or ` apt list --installed`, to confirm the packages which might be mising from the list. :)

PR #512, opened 2021-10-14
Schema YML Configuration

prdanelli:
Refactor work type field definitions to be schema driven.

@cjcolvar Hyrax PR: https://github.com/samvera/hyrax/pull/5028

When reviewing the PR, you will need to ignore the majority of the view partials as they have small changes, mostly including `if f.object.schema_driven?` at the top to preserve backwards compatibility. 

To Ignore: 
- Most simple field related views 
  - alternate_identifier/_edit_array_hash_form.html.erb 
  - alternate_identifier/_new_form.html.erb
  - current_he_institution/_edit_array_hash_form.html.erb
  - etc...
- Schema related modules that were copied from Hyrax 3.x 
  - app/presenters/hyku_addons/schema/presenter.rb 
  - lib/hyrax/indexer.rb
  - lib/hyrax/schema_generator.rb
    - schema generator yaml partials found in config/metadata/partials.
  - lib/hyrax/simple_schema_loader.rb 
- CSS

To review: 
- Readme changes
- Changes to actors
- All JS files
- WorkForm / WorkBase concerns 
  -  app/forms/concerns/hyku_addons/schema/work_form.rb
  - app/forms/concerns/hyku_addons/work_form.rb
  - app/models/concerns/hyku_addons/schema/work_base.rb
  -  app/models/concerns/hyku_addons/work_base.rb
- Any "customer" specific works included - anything related to UbiquityTemplateWork or UvaWork
  - Form
  - Presenter
  - Model
  - Indexer
  - Controllers
- SimpleForm Input classes
- Hyrax::FormFields - lib/hyrax/form_fields.rb, contains HykuAddons specific logic and is customised from Hyrax
- Specs / fixtures / helpers
- Yaml schema in config/metadata


app/actors/hyku_addons/actors/note_field_actor.rb
```diff
@@ -15,14 +15,15 @@ def update(env)
       private
 
         def serialize_note_field(env)
+          return unless env.curation_concern.respond_to?(:note)
+
           if env.attributes[:note].present?
```
thenapking, line 20: I am assuming here that `env.curation_concern.note` is an array, as it has another array appended in this code.  Given that a more terse way of writing this might be:
````
def serialize_note_field(env)
  return unless env.curation_concern.respond_to?(:note)
  
  @email = env.user.email if env.attributes[:note].present?
    
  env.attributes[:note] = note_array(env).push(*process_note_hash(env))
end

def note_array(env)
  env.curation_concern.note || []
end

def process_note_hash(env)
  return unless env.attributes[:note]

  Array(
    { email: email, timestamp: Time.zone.now.to_s, note: env.attributes[:note] }.to_json
  )
end
````

This is untested though and may contain typos or other errors!  It's also possible to refactor this down more if `@email` is only being set here for use in `process_note_hash`.  NB This will return `[]` not `nil` if `env.curation_concern.note.nil?` and `env.attributes[:note].nil?` which might not be what you want.

Comment (prdanelli) [reply app/actors/hyku_addons/actors/note_field_actor.rb:20]: Thanks for this @thenapking. The only line I actually needed was here https://github.com/ubiquitypress/hyku_addons/pull/512/files/97c0d1ef8ee8f06c210594dd93d248445ea12ed7#diff-f4cd9cb2f565f29a5825dbcd96ca79b2082a0871a4f39cef36f4d8bea91699d6R18

Which was just to ensure that the save process wouldn't fail when a note field was missed. The rest is just styling the variable assignment to be indented and easier to read. 

If you want to put in a PR with the changes above then that would be great :)

File: app/views/hyku_addons/dates/_date_submitted.html.erb [hunk 2]
```diff
@@ -1,8 +1,2 @@
-<%# see app/views/records/edit_fields/_date_submitted.html.erb %>
-<% template = f.object.model.class.to_s.underscore %>
-
-<%= render 'hyku_addons/dates/date_fields', date_type: 'submitted', template: template, f: f %>
-
-
-
-<br/><br/><br/>
+<!-- legacy - _date_submitted -->
+<%= render "hyku_addons/dates/date_fields", date_type: "submitted", template: f.object.model.class.to_s.underscore, f: f %>
```
thenapking, line 2: This is a good refactor to remove the `template` variable which is defined in this partial. But now all this partial is doing is setting the value for `date_type` and opening another partial.  The other partial _date_published and _date_accepted just set different date_types.  Can't we refactor somewhere further up the chain so `data_type` is set there and have only one partial not three?

prdanelli: By default the hydra editor gem uses the `key` (underscored field name) to work out the partial to include. Because of the way that the dates are processed - by creating a JSON string of date components year, month, day, which is passed to a date processing actor and combined into a YYYY-MM-DD string - we cannot use the `_default` partial which is the last file called because a 404 is thrown. The `_default` is where I have put most of the logic for rendering the views, based on the schema definitions, but dates need more time than I currently have. 

When this work is done, then changing the way the dates are processed would be of real benefit and simply the stack - just using Rails `date(2i)` internal processing instead.

spec/helpers/fill_in_fields.rb
```diff
@@ -0,0 +1,245 @@
+# frozen_string_literal: true
+
+# Fillin the form files
```
thenapking, line 3: `spec/helpers` should only contain tests for helper classes.  Helpers for the test suite should be in `spec/support`.

thenapking (approved):

app/actors/hyku_addons/actors/json_fields_actor.rb
```diff
@@ -14,18 +14,22 @@ def update(env)
...
         def jsonify_fields(env)
           env.curation_concern.class.json_fields.each do |field|
+            # This handles the case when field is a key/value pair coming from the yaml schema
+            field = field.first if field.is_a?(Array)
+
             if name_blank?(field, env.attributes[field]) || recursive_blank?(env.attributes[field])
               env.attributes.delete(field)
             else
-              env.attributes[field].reject! { |o| name_blank?(field, o) || recursive_blank?(o) } if env.attributes[field].is_a? Array
+              env.attributes[field].reject! { |o| name_blank?(field, o) || recursive_blank?(o) } if env.attributes[field].is_a?(Array)
```
thenapking, line 23: You define `name_blank?` and `recursive_blank?` identically in two places.  Also when it's used it's always `name_blank? || recursive_blank?`.  Could you move both definitions to a module, and create one method which will test the full condition?

app/forms/concerns/hyku_addons/schema/work_form.rb
```diff
@@ -0,0 +1,79 @@
...
+      included do
+        class_attribute :primary_fields, :field_configs, :internal_terms
+        self.primary_fields = []
+        self.required_fields = []
+        self.field_configs = {}
+        self.internal_terms = [] # Internal hyrax terms which don't require form fields
+
+        def self.build_permitted_params
+          super.tap do |permitted_params|
+            model_class.json_fields.deep_symbolize_keys.each do |field, field_config|
```
thenapking, line 16: Perhaps move this bit into a private method to reduce the cognitive complexity?

app/actors/hyku_addons/actors/json_fields_actor.rb
prdanelli: This is part of the original json actor that was inherited from the OG PR. I have tried not to touch as much as possible, but you're probably right that extra work might want to be done to simplify it here. I would suggest its best to leave that for now, as we know it works currently and so it could be discounted if errors start to pop up.

spec/features/ubiquity_template_work_form_spec.rb
```diff
@@ -0,0 +1,505 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+require HykuAddons::Engine.root.join("spec", "support", "fill_in_fields.rb").to_s
+require HykuAddons::Engine.root.join("spec", "support", "work_form_helpers.rb").to_s
+
+RSpec.feature "Create a UbiquityTemplateWork", js: true do
```
thenapking, line 7: Should this be labelled as a slow test?

spec/features/uva_work_form_spec.rb
```diff
@@ -0,0 +1,187 @@
...
+  let(:workflow) { Sipity::Workflow.create!(active: true, name: "test-workflow", permission_template: permission_template) }
+  let(:new_work_path) { "concern/#{work_type.to_s.pluralize}/new" }
+  let(:permission_options) do
+    { permission_template_id: permission_template.id, agent_type: "user", agent_id: user.user_key, access: "deposit" }
+  end
+  # The organisation option changes depending on the local, so we need to use this to ensure we select the right one
+  let(:organisation_option) { HykuAddons::NameTypeService.new(model: model).active_elements.last }
+
+  let(:title) { "UVA Work Item" }
+  let(:creator) do
```
thenapking, line 24: Both this an the `ubiquity_template_work_form_spec` share a lot of definitions and expectations.  Is there any sensible way to bundle these up into groups and share them?

cjcolvar:
Congrats on getting this merged! :tada: :fireworks: :confetti_ball:

prdanelli:
@cjcolvar Thank you for the work you put in to it. 

There are still some bugs i'm finding, but only in production (AH).

PR #518, opened 2021-10-19
Redirect from shared-search result to the right tenant

mankind:


app/presenters/concerns/hyku_addons/collection_presenter_override.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+module HykuAddons
+  module CollectionPresenterOverride
+    extend ActiveSupport::Concern
+
+    # override by calling elevator switch! in shared_search result it raises error Record not found error
+    def collection_type
+      work_cname = @solr_document.to_h["account_cname_tesim"]
+      account_cname = work_cname.class == Array ? work_cname.try(:first) : work_cname
```
prdanelli, line 9: Could we use `Array.wrap(work_cname).first` instead of the type checking?

spec/helpers/cross_tenant_shared_search_helper_spec.rb
```diff
@@ -15,30 +15,75 @@
     context '#feature_for_display_in_proprietor_account_ui' do
       let(:proprietor_account_flipflop_features_list) { helper.feature_for_display_in_proprietor_account_ui }
 
-      it 'returns truthy' do
+      it "returns truthy" do
```
prdanelli, line 18: LOVE THIS!

prdanelli (changes requested):
It looks great, just a few comments on tidying up a couple of bits. :)

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
```diff
@@ -3,6 +3,27 @@
...
+      has_model = model["has_model_ssim"].first.underscore.pluralize
+      # returns a symbol not a string
+      id = model[:id]
+      request_params = { request_protocol: request_protocol, request_host: request_host, request_port: request_port }
+
+      if has_model == "collections"
+        set_collection_url(id: id, request: request_params, account_cname: account_cname, has_model: has_model)
+      else
+        set_work_url(id: id, request: request_params, account_cname: account_cname, has_model: has_model)
+      end
```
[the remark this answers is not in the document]
prdanelli: There is just a lot of duplication, which adds mental overhead to the method. You're not actually using `request_host` etc variables, so you could just add `request.host` to the hash instead. But reduicing the method down from 19 lines to: 

```
    def generate_work_url(model, request)
       account_cname = Array.wrap(model["account_cname_tesim"]).first
       has_model = model["has_model_ssim"].first.underscore.pluralize
       request_params = %i[protocol host port].map { |method| ["request_#{method}".to_sym, request.send(method)] }.to_h

        method_name = has_model == "collection" ? :set_collection_url : set_work_url
        send(method_name, id: model["id"], request: request_params, account_cname: account_cname, has_model: has_model)
    end
```

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
```diff
@@ -4,28 +4,23 @@
 module HykuAddons
   module CrossTenantSharedSearchHelper
     def generate_work_url(model, request)
+      # returns a symbol for id and string for other attributes
       model = model.with_indifferent_access
```
prdanelli, line 8: It look like you have this here so that you can use `model[:id]`, but I can't see anywhere else where it would be used. Could we remove it and then access the ID via `model["id"]`?

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
```diff
@@ -4,28 +4,23 @@
...
       id = model[:id]
-      request_params = { request_protocol: request_protocol, request_host: request_host, request_port: request_port }
 
-      if has_model == "collections"
-        set_collection_url(id: id, request: request_params, account_cname: account_cname, has_model: has_model)
-      else
-        set_work_url(id: id, request: request_params, account_cname: account_cname, has_model: has_model)
-      end
+      request_params = %i[protocol host port].map { |method| ["request_#{method}".to_sym, request.send(method)] }.to_h
+      set_url(id: id, request: request_params, account_cname: account_cname, has_model: has_model)
```
prdanelli, line 16: This is now the only place where `id` is being referenced, so we could replace this with `model["id"]` and save the variable assignment.

prdanelli (changes requested):
Just left a couple of comments for that one method. Otherwise its great :)

app/helpers/hyku_addons/cross_tenant_shared_search_helper.rb
mankind: I can remove the varibale assignment but  **model[:id]** will stay as it is,  because I discovered the hash from solr returns **id** as symbol and the other attributes we are using here eg **has_model_ssim** as string.

PR #519, opened 2021-10-19
Moves exports to background jobs

okeen:


app/controllers/concerns/hyku_addons/exporters_controller_override.rb
```diff
@@ -0,0 +1,38 @@
...
+          else
+            render :new
+          end
+        end
+
+        def update
+          field_mapping_params
+          if @exporter.update(exporter_params)
+            # Use perform later for creating the export in the background
+            Bulkrax::MultitenantExporterJob.perform_later(current_account.id, @exporter.id) if params[:commit] == 'Update and Re-Export All Items'
```
prdanelli, line 29: This seems really brittal. Is there another way to test if its an update?

prdanelli (approved):

PR #520, opened 2021-10-20
Bixby Rubocop issues

prdanelli:
This might not affect everyone, however the version of Rubocop we are using is very old (2018) because the version of bixby we are using is stuck on 1.0.0 by hyrax 2.9.x. 

This is causing issues for the Ruby language server Solargraph, which requires a very old version to be compatible with the older versions of Bixby/Rubocop.


Gemfile
```diff
@@ -22,4 +22,19 @@ gemspec
...
+# Try and update Rubocop/bixby and related Gems locks in 2018
+gem 'rubocop', require: false
+gem 'rubocop-performance', require: false
+gem 'rubocop-rails', require: false
+gem 'rubocop-rspec', require: false
+
+file_name = File.expand_path('spec/internal_test_hyku/Gemfile', File.dirname(__FILE__))
+text = File.read(file_name)
+new_contents = text.gsub(/gem 'rubocop', '~> 0.50', '<= 0.52.1'/, "")
+                   .gsub(/gem 'rubocop-rspec', '~> 1.22', '<= 1.22.2'/, "")
```
okeen, line 34: You are mixing single and double quotes here

okeen (approved):
Looks OK, just a small note

PR #521, opened 2021-10-20
rake task and sidekiq job to reindex model

mankind:


app/jobs/hyku_addons/reindex_available_works_job.rb
```diff
@@ -0,0 +1,25 @@
...
+      def fetch_work_save_cname(cname)
+        AccountElevator.switch!(cname)
+        available_works = Site.first.available_works | ['Collection']
+
+        available_works.each do |model_class|
+          klass = model_class.constantize
+          contains_work = klass.count.positive?
+
+          Rails.logger.debug "==== Queue ReindexModelJob for #{klass} ===" if contains_work
+          HykuAddons::ReindexModelJob.perform_now(model_class, cname) if contains_work
```
prdanelli, line 21: This is the part that worries me, as it'll block other jobs for the time it takes to reindex the entirity of Fedora. THen if you get an error that you're not catching, it'll requeue its self and its back to the start again. 

It would be better to have multiple small jobs, which were queued when the first job finished. 

I was thinking, something like this: 

```module HykuAddons
  class ReindexAvailableWorksJob < ApplicationJob
    def perform(cnames)
      # Shift the first item off the array and store it
      cname = cnames.shift

      # If cname is nil, we have run out of cnames to process
      return if cname.blank?

      reindex(cname)

      # Requeue the cnames without the one just processed
      HykuAddons::ReindexAvailableWorksJob.perform_later(cnames)
      # Or maybe this could be... 
      # self.perform_later(cnames) 
    end

    def reindex(cname)
      AccountElevator.switch!(cname)

      available_works = Site.first.available_works | ['Collection']

      available_works.each do |model_class|
        klass = model_class.constantize
        contains_work = klass.count.positive?

        next unless contains_work?

        Rails.logger.debug "=== Starting to reindex #{klass} in #{cname} ==="

        klass.constantize.reindex_everything

        Rails.logger.debug "=== Completed reindex of #{klass} in #{cname} ==="
      end
    end
  end
end
```
This way we only have a single job, we don't potentially block the processing of other queues for very long and when a job completes, it is requeued, so other jobs are able to be processed and then it will run and requeue the next version of its self.... recursively.

app/jobs/hyku_addons/reindex_available_works_job.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+module HykuAddons
+  class ReindexAvailableWorksJob < ApplicationJob
+    def perform(array_of_cname)
```
prdanelli, line 4: We can just call this `cnames`, so its a litte less cumbersome, and as its a single line block, could be moved up. 

```
def perform(cnames)
  cnames.each { |cname| fetch_work_save_cname(cname) }
end
```

mankind: I want someone reading to know by just looking at it. Naming in a way that the intent or what is expected is obvious seems easier for clarity even if the other is shorted.

app/jobs/hyku_addons/reindex_available_works_job.rb
mankind: I @okeen approach because if we add rescue from error then any error will only affect stop that one model from being re_indexed. 
My old approach before the standup and what you are suggesting seems similar. My concern was that in that case when an error occurs, it stops re-Indexing for the whole models in that tenant.
I think Article should be able to re_index even if GenericWork fails and that is not the case with my old approach and what you are suggesting. 
That was the concern I had that led to our discussion this morning.

app/jobs/hyku_addons/reindex_available_works_job.rb
prdanelli: I would provide an example of how to use the task in the rake file, so that the syntax and format do not have to be worked out by the reader, like so many of the hyku/hyrax/Addons tasks.

mankind: We already have that comment in the rake file pushed with this PR.  However, if you open just the job file, you should  know what is expected without going to the rake task, since it is not unfathomable to think of scenarios that someone wants to use the job without the rake.

We can achieve that clarity by either adding a comment or naming things properly. 

I will change the name though and add comments to the job since the rake file already has that comment.

app/jobs/hyku_addons/reindex_available_works_job.rb
mankind: I think the PR in its current form also does this too:

- It's not one job processing these items, but multiple smaller jobs, which is what background jobs should be - just doing one thing and then moving one. **You are creating multiple job of same class and the PR does same but with 2 classes**
- Handles the one or more cnames passed into the array in the same way, rather than needing to have the arguments as a single cname for each time the rake task is called. 

So the only difference between the PR and your suggestion is point 1:
- If there is an error, it will requeue the job with the outstanding cnames array intact and we do not have to start again.

We can add the removal from the array to this PR and that should bring us to thesame point

prdanelli (approved):

PR #522, opened 2021-10-20
UNA Presentation UNA Thesis and UNA Time Based Media

BertZZ:


app/views/hyrax/una_thesis_or_dissertations/_una_thesis_or_dissertation.html.erb
```diff
@@ -0,0 +1,2 @@
+<%# This is a search result view %>
+<%= render 'catalog/document', document: una_thesis_or_dissertation, document_counter: una_thesis_or_dissertation_counter  %>
```
prdanelli, line 2: You're using single quotes here and double quotes for the other view partial. Can they all be doubles please.

prdanelli (changes requested):
SO close. :)

PR #533, opened 2021-11-16
Hypothesis integration

thenapking:
On a work’s page, for each fileset the user can download, add a link to the Actions dropdown, which opens the file using hypothesis external service.

### Work completed
- [ ] Add Flipflop config control for `:annotation`
- [ ] Override [views/hyrax/file_sets/actions](https://github.com/samvera/hyrax/blob/main/app/views/hyrax/file_sets/_actions.html.erb) partial
- [ ] Refactor logic to test for config control.  When this is switched on, users with only the download permission will now be able to see two links and so the view must also display ul for them.
NB I have removed the configuration to run this locally as I can see little need to do this as we are essentially linking out to a third party service which we will not interact with afaik.

### Layout
image

### Ticket
https://ubiquitypress.atlassian.net/browse/REPO-1495

### DIFF
--- a/app/views/hyrax/file_sets/_actions.html.erb
+++ b/app/views/hyrax/file_sets/_actions.html.erb
@@ -1,0 +1,58 @@
+<% if can?(:edit, file_set.id) || can?(:destroy, file_set.id) %>
+  <div class="btn-group">
+    <button class="btn btn-default dropdown-toggle" data-toggle="dropdown" type="button" id="dropdownMenu_<%= file_set.id %>" aria-haspopup="true" aria-expanded="false">
+      <span class="sr-only"><%= t('.press_to') %> </span>
+      <%= t('.header') %>
+      <span class="caret" aria-hidden="true"></span>
+    </button>
+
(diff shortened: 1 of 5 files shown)


prdanelli (commented):
This is great Tom. I just have one request. There are a mix of single and double quotes in the specs and single seems to be being used just for renders: `render 'hyrax/file_sets/actions', file_set: file_set`. Can those be double quotes please.

thenapking:
No problems, I must have missed that.  Do you think this is something we could alter the rubocop definition to meet so this happens automatically?

mankind:
> No problems, I must have missed that. Do you think this is something we could alter the rubocop definition to meet so this happens automatically?

I think I will allow @prdanelli  7 @BertZZ to chime in.

thenapking:
I have updated the rubocop config so that it enforces double quotes in the test suite.  Obviously this has touched a lot of files, so I have put this into another PR: https://github.com/ubiquitypress/hyku_addons/pull/534

thenapking:
Double quotes fixed in: [56efcb5](https://github.com/ubiquitypress/hyku_addons/pull/533/commits/56efcb54e17fa881181f75f5c074e1592d8fd53e).

thenapking:
Closing this pull request as another one is in development.

PR #534, opened 2021-11-16
Update Rubocop config to enforce double quotes in test suite

thenapking:
Your dream/wish/nightmare has come true.

Update rubocop config to enforce double quotes 
For safety, I have applied this only to the test suite.  All tests updated tests still pass.


prdanelli (approved):
If the specs all pass, this is great, thank you. Also an amazing way to get your OS contributions up :D

PR #536, opened 2021-11-17
Flaky Hyrax::DOI::RegisterDOIJob fix?

thenapking:
`Hyrax::DOI::RegisterDOIJob` regularly fails because the `cname` attribute of `account` has not been set.  Since we are not particularly concerned about this being randomised for this test, we can fix it first and create an account with a matching `cname`.  Then the `cname` will always have a value and this test hopefully will stop failing.

### Work completed
- [ ] Remove double definition of account
- [ ] Fix cname value
- [ ] Refactor test for brevity
- [ ] Refactor factories

### DIFF
--- a/spec/factories/datacite_endpoint.rb
+++ b/spec/factories/datacite_endpoint.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :datacite_endpoint do
+    options { Hash.new(mode: "test", prefix: "10.1234", username: "user123", password: "pass123") }
+  end
+end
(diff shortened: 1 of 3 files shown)


spec/factories/datacite_endpoint.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
```
prdanelli, line 2: Can I just say, I'm a massive fan of people who leave a space below the `#frozen_string_literal`. ❤️

prdanelli (approved):
Fingers crossed! :)

PR #537, opened 2021-11-18
Update docker compose file to properly share configuration

prdanelli:


Dockerfile
```diff
@@ -44,10 +44,10 @@ COPY --chown=app:app spec/internal_test_hyku/Gemfile ./spec/internal_test_hyku/G
...
-RUN bundle config build.nokogiri --use-system-libraries \
-	&& bundle config set without 'production' \
-	&& bundle config set with 'aws development test postgres' \
-	&& setuser app bundle install --jobs=4 --retry=3
+RUN bundle config build.nokogiri --use-system-libraries && \
+    bundle config set without "production" && \
+    bundle config set with "aws development test postgres" && \
+    setuser app bundle install --jobs=4 --retry=3
 
 RUN chmod 777 .bundle/config # Otherwise `app` owns this file and the host cannot run bundler commands
```
mankind, line 52: As we saw with my setup during hyrax_autopopulation  gem, we still could not run bundler outside docker with doing another **chmod 777** in the host mchine, do you want to try this instead 

      RUN chmod 777  -R .bundle/*

mankind (approved):
:+1:

PR #539, opened 2021-11-18
Switches canonical identifier

BertZZ:


spec/lib/work_factory_spec.rb
```diff
@@ -19,14 +19,14 @@
       ],
       "uri": [
```
prdanelli, line 20: You have a new strings being converted to symbol here that dont need to be wrapped in speech marks. Could you remove the quotes and just use symbols :)

prdanelli (changes requested):

prdanelli (approved):

PR #540, opened 2021-11-18
Fix account behaviour spec failure

thenapking:
This is a possible fix for the flaky account_behaviour_spec.

In this commit https://github.com/ubiquitypress/hyku_addons/commit/ce74690d713d3e955535b63b515c7aefdfd523a1, the behaviour of `.reset!` was changed so that the mode is reset to `:test` rather than `nil`.  When `switch` (not `switch!` with a bang) is called and fails, then `reset!` is called (see source code [here](https://github.com/samvera/hyku/blob/v.2.1.0/app/models/account.rb#L94-L105)).  Hence the value of mode we should test for in the `it "resets the active connections back to the defaults"` test will always be `:test` not `nil`.  

Also it is worth noting that the config specifies the two valid values of mode are symbols not strings, but the test suite uses strings to setup the test objects, and so I have standardised this.

### DIFF
--- a/spec/factories/datacite_endpoint.rb
+++ b/spec/factories/datacite_endpoint.rb
@@ -2,6 +2,6 @@
 
 FactoryBot.define do
   factory :datacite_endpoint do
-    options { Hash.new(mode: "test", prefix: "10.1234", username: "user123", password: "pass123") }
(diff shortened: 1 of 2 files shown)


prdanelli (approved):
Thanks for looking at this. If the specs all pass then lets merge it in.

PR #542, opened 2021-11-22
Denver Select Work Type Model Changes

prdanelli:


config/locales/en-DENVER.yml
```diff
@@ -58,8 +93,6 @@ en-DENVER:
         label: Role
       contributor_grid:
         label: GRID
-      contributor_role:
-        label: Role
```
BertZZ, line 62: Any reason for removing these lines??

prdanelli: Ok, thats super wierd. My editor flagged them as duplicate, so i removed the extra one, but now its saying that there is only one. 🤔

prdanelli: This is what i'm seeing. 
image

BertZZ (approved):

PR #548, opened 2021-11-26
Add new file set registration job to the bulkrax before do blocks

BertZZ:


BertZZ:
Not needed as specs were fixed

PR #549, opened 2021-11-29
Fix Bulkrax Test Failures

thenapking:
After an update to the Hyrax Hirmeos tests relating to Bulkrax were failing.  There may have been no issue with the Hyrax Hirmeos update, but because the Bulkrax test ran *all* jobs, it is entirely likely the Hirmeos ones failed because the test was not set up with the information required for those jobs to run.

The failing tests tested whether files were imported as part of a feature test which involves running two separate jobs.  Because the original author of this test could not get a special configuration feature to work, code was added to run all jobs in the queue.  This is bad practice, as jobs should be run in isolation.  This pattern is repeated throughout the test suite and needs to be removed.

### Work done
Remove unnecessary use of workaround to filter jobs and adjust queue adaptor
Add in block which specifies which related jobs should be run

### DIFF
--- a/spec/features/bulkrax_export_spec.rb
+++ b/spec/features/bulkrax_export_spec.rb
@@ -2,7 +2,7 @@
 
 require "rails_helper"
 
-RSpec.describe "Bulkrax export", clean: true, perform_enqueued: true do
+RSpec.describe "Bulkrax export", clean: true do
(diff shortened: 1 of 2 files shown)


prdanelli (approved):

prdanelli:
Thanks @thenapking this is great!

PR #552, opened 2021-11-30
switch find_by with activerecord exists?

mankind:
Performance improvement because using Benchmark.measure to compare **Account.all.map(&:cname)** and **Account.exists?(cname: account_cname)** showed the latter was about 2o times faster

### DIFF
--- a/app/presenters/concerns/hyku_addons/collection_presenter_override.rb
+++ b/app/presenters/concerns/hyku_addons/collection_presenter_override.rb
@@ -7,8 +7,7 @@ module HykuAddons
     def collection_type
       work_cname = @solr_document.to_h["account_cname_tesim"]
       account_cname = Array.wrap(work_cname).first
-      @check_cname_exists = Account.exists?(cname: account_cname)
-      return super unless @check_cname_exists.present?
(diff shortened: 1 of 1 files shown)


app/presenters/concerns/hyku_addons/collection_presenter_override.rb
```diff
@@ -7,8 +7,8 @@ module CollectionPresenterOverride
     def collection_type
       work_cname = @solr_document.to_h["account_cname_tesim"]
       account_cname = Array.wrap(work_cname).first
-      @all_cname = Account.all.map(&:cname)
-      return super unless @all_cname.include? account_cname
+      @check_cname_exists = Account.exists?(cname: account_cname)
+      return super unless @check_cname_exists.present?
```
thenapking, line 11: Why do we need both `.exists?` and `.present?`.  Can't it just be:
````
return super unless Account.exists?(cname: account_cname)
````

mankind: Thanks I change that

prdanelli (approved):

PR #554, opened 2021-11-30
Ensure each test only processes relevant jobs

thenapking:
Following on from the bug fix in [549](https://github.com/ubiquitypress/hyku_addons/pull/549), by applying the same refactoring to each place the pattern was present in the test suite.
Previously these tests processed *all* jobs that got enqueued.  Now, either test that the required job is enqueued with the correct parameter (and rely on the tests for that job to know the application will end up in the correct state) or run only those jobs required for the feature.

### Work completed
Remove `config.around(:example, :perform_enqueued)` config block
Ensure all outstanding tests are cleared at the end of each run
Refactor HyraxDoiSpec by splitting 'it mints a DOI' into three tests.  
-First test checks the user is redirected after the form is filled in.  
-Second test checks that the appropriate jobs are logged with the right parameters.
-Third test checks that if an existing work has a minted DOI then the link to it is displayed on its show page.
Refactor `ReindexAvailableWorksJob` and `ReindexModelJob` to take account of the fact that all jobs will not run
-These jobs both spawn more `ReindexModelJob`s.  So when `.perform_now` is run we should see instances of `ReindexModelJob` enqueued, and once instance of the spawning job performed.

@mankind could you please review this to confirm that I have understood the `ReindexAvailableWorksJob` and `ReindexModelJob` classes correctly.

### DIFF
--- a/spec/features/hyrax_doi_spec.rb
+++ b/spec/features/hyrax_doi_spec.rb
@@ -130,60 +130,29 @@ RSpec.describe "Minting a DOI for an existing work", multitenant: true, js: true
 
     context "when the user selects `findable`" do
       let(:new_title) { "New work title" }
-      let(:doi) { "#{prefix}/kgkc-nn31" }
 
-      let(:response_fixture) { File.read Rails.root.join("..", "fixtures", "datacite", "put_metadata.xml") }
-      let(:common_headers) do
-        {
-          "Accept": "*/*",
-          "Accept-Encoding": "gzip;q=1.0,deflate;q=0.6,identity;q=0.3",
-          "Authorization": "Basic VkpLQS5KQ1JYWkctTE9DQUw6cGFzc3dvcmQ=",
-          "User-Agent": "Faraday v0.17.4"
-        }
-      end
-      let(:json_headers) do
-        common_headers.merge("Content-Type": "application/vnd.api+json")
-      end
-      let(:xml_headers) do
-        common_headers.merge("Content-Type": "application/xml;charset=UTF-8")
-      end
-      let(:text_headers) do
-        common_headers.merge("Content-Type": "text/plain;charset=UTF-8")
+      it "redirects the user to the works page" do
+        fill_in_form(new_title)
+
+        expect(page).to have_selector("h1", text: "Work", wait: 10)
+        expect(page).to have_selector("h2", text: new_title)
       end
 
-      before do
-        # The initial request to create_draft_doi
-        stub_request(:post, "https://api.test.datacite.org/dois")
-          .with(body: { "data": { "type": "dois", "attributes": { "prefix": prefix } } }.to_json, headers: json_headers)
-          .to_return(status: 201, body: { "data": { "id": doi, "type": "dois", "attributes": {} } }.to_json, headers: {})
-
-        # Send the work data to datacite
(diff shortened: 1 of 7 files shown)


spec/features/hyrax_doi_spec.rb
```diff
@@ -130,60 +130,29 @@ def work_url(work)
...
-        stub_request(:put, "https://mds.test.datacite.org/metadata/#{doi}")
-          .with(body: response_fixture, headers: xml_headers)
-          .to_return(status: 201, body: "", headers: {})
-
-        # Register the URL
-        stub_request(:put, "https://mds.test.datacite.org/doi/#{doi}")
-          .with(body: "doi=#{doi}\nurl=http://#{cname}/concern/generic_works/#{work.id}", headers: text_headers)
-          .to_return(status: 201, body: "", headers: {})
+      it "enqueues a job to mint the DOI" do
+        expect { fill_in_form(new_title) }.to have_enqueued_job(Hyrax::DOI::RegisterDOIJob).with(work.reload, registrar: "datacite", registrar_opts: {})
```
mankind, line 142: I think from the UI, just filling a title should not cause a doi to be minted. I think it is checkbox you need to check.

See attach image. So we need to check draft
doi

mankind (approved):
Looks good :+1: 

Not an issue  but sometimes you are calling exactly(:once) and other times at_least(:once). Also for one perform_later test it did not have any of exactly once or at_least(:once)

thenapking:
> Looks good 👍
> 
> Not an issue but sometimes you are calling exactly(:once) and other times at_least(:once). Also for one perform_later test it did not have any of exactly once or at_least(:once)

That's correct.  On `ReindexModelJob` I call `exactly(:once)` because at the end of that job it re-enqueues itself once.  `ReindexAvailableWorksJob` enqueues `ReindexModelJob` for every available work, but I could not find a reliable way to count these, and so I opted for `at_least(:once)`, but I am happy to refactor if you can point me in the right direction.  Both calls to `.perform_later` should just enqueue the job in question.  I wouldn't normally test this... it appears to be testing the functionality of `.perform_later` rather than the job itself.  But I didn't want to remove this in case there was some monkey-patching, or another reason I didn't know about for testing this.

mankind:
> > Looks good +1
> > Not an issue but sometimes you are calling exactly(:once) and other times at_least(:once). Also for one perform_later test it did not have any of exactly once or at_least(:once)
> 
> That's correct. On `ReindexModelJob` I call `exactly(:once)` because at the end of that job it re-enqueues itself once. `ReindexAvailableWorksJob` enqueues `ReindexModelJob` for every available work, but I could not find a reliable way to count these, and so I opted for `at_least(:once)`, but I am happy to refactor if you can point me in the right direction. Both calls to `.perform_later` should just enqueue the job in question. I wouldn't normally test this... it appears to be testing the functionality of `.perform_later` rather than the job itself. But I didn't want to remove this in case there was some monkey-patching, or another reason I didn't know about for testing this.

Thanks @thenapking , I think we can leave it as it is for now and I agree that we should not test perform_later or any method provided by **Rails**. We can refactor it to test  it enqueues with the correct parameters which should be easier than testing  the outcome of triggering the job. 
I have nothing against removing it though my preference will be to test it enqueues with the right parameters.

prdanelli (approved):

PR #558, opened 2021-12-01
ImageMagick gs permissions spec

prdanelli:
Adding a spec to confirm that permission issues have been resolved.


prdanelli:
Not required anymore as we have fixed the issue by updating ImageMagick to 7.x.

PR #560, opened 2021-12-02
Api should return cname as a string

mankind:
The Api is returning **cname** as an array which is a breaking change that makes the React frontend not to be able to show links in shared search result. This PR returns the cname as a string.


app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -18,7 +18,7 @@ json.cache! [@account, :works, work.id, work.solr_document[:_version_], work.mem
   json.buy_book work.try(:solr_document)&.to_h&.dig('buy_book_tesim')
   json.challenged work.try(:solr_document)&.to_h&.dig('challenged_tesim')
   json.citation work.try(:solr_document)&.to_h&.dig('citation_tesim')
-  json.cname @account.search_only? ? work.try(:solr_document)&.to_h&.dig("account_cname_tesim") : @account.cname
+  json.cname @account.search_only? ? work.try(:solr_document)&.to_h&.dig("account_cname_tesim").first : @account.cname
```
prdanelli, line 21: You will need to use `&.first` here

mankind: Thanks added and pushed.

prdanelli (approved):

PR #564, opened 2021-12-03
Add task to regenerate work pdf thumbnails

prdanelli:


mankind:
Looks good :+1: . Do you know if there are any collection's whose thumbnail has the issue and needs regeneration?

mankind (dismissed):

lib/tasks/hyku_addons/file_set.rake
```diff
@@ -0,0 +1,31 @@
...
+        count = work_type.count
+        per_page = 25
+        pages = (count.to_f / per_page.to_f).ceil
+
+        pages.times do |page|
+          work_type.limit(per_page).offset(page * per_page).each do |work|
+            work.file_sets.each do |file_set|
+              next unless file_set.mime_type.match?(/^(.)+\/pdf$/)
+
+              CreateDerivativesJob.set(wait: rand(3600)).perform_later(file_set, file_set.original_file.id)
```
thenapking, line 22: I presume you are doing this so as not to overload the job queue and spread the jobs out.  Wouldn't it be better to have a job which spawns the CreateDerivatesJobs for a fixed number of the works which need updating  every hour.  So we'd update say 25 works an hour, and let the workers deal with managing their own loads rather than spreading it out randomly.

prdanelli: I did think about it, however simplicity seemed preferable as this might never be used again.

PR #565, opened 2021-12-06
REPO 1495 add link to hypothesis pdf js

thenapking:
On a work’s page, for each fileset the user can download, add a link to the Actions dropdown, which opens the file using hypothesis viewer.  Note that this PR looks large because it brings in all the Hypothes.is and Mozilla assets.  You do not need to review files in /assets, with the exception of pdf-viewer.js

### Work completed
- [x] Bring in hypothes.is assets to the asset pipeline
- [x] Create new controller to serve assets and show PDF viewer
- [x] Update hypothes.is JS to prevent showing of random files / malicious usage
- [x] Add Flipflop config control for `:annotation`
- [x] Override [views/hyrax/file_sets/actions](https://github.com/samvera/hyrax/blob/main/app/views/hyrax/file_sets/_actions.html.erb) partial
- [x] Refactor logic to test for config control.  When this is switched on, users with only the download permission will now be able to see two links and so the view must also display ul for them.

### Layout
image

### Ticket
https://ubiquitypress.atlassian.net/browse/REPO-1495


spec/requests/hyku_addons/pdf_viewer_controller_request_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+
+  after do
+    account.reset!
+  end
+
+  describe "GET pdf_viewer" do
+    it "loads hypothesis pdf viewer" do
+      get hyku_addons.pdf_viewer_path(pdf_path)
+      expect(response).to have_http_status(:ok)
+      expect(response.body).to have_css("#outerContainer") # outer container of the hypothes.is html
```
mankind, line 32: So if the css id is present is guaranteed to have loaded the pdf, are there any possibilities of  having the css id  in the body without pdf being loaded?

app/assets/javascripts/pdf_viewer/base.js
```diff
@@ -0,0 +1 @@
+//= require_tree .
```
mankind, line 1: Just checking does this not load everything in the JavaScript directory? 
If so is having it here different from having it in **app/assets/javascripts/hyku_addons.js**

app/assets/javascripts/pdf_viewer/web/pdfjs-init.js
```diff
@@ -0,0 +1,26 @@
...
+// nb. After making changes to this file, the copy in the viewer/web/ dir
+// must be updated.
+
+// Listen for `webviewerloaded` event to configure the viewer after its files
+// have been loaded but before it is initialized.
+//
+// PDF.js >= v2.10.377 fires this event at the parent document if it is embedded
+// in a same-origin iframe. See https://github.com/mozilla/pdf.js/pull/11837.
+try {
+  parent.document.addEventListener('webviewerloaded', onViewerLoaded);
```
mankind, line 12: Again this is just a question or check. I know your layout does not load turbolinks but seeing that you **require_tree .** the javascript directory in **app/assets/javascripts/pdf_viewer/base.js** do we need to wrap this call with the line below  or you are sure turbolinks is not loaded when **require _tree  .**:

      $( document ).on('turbolinks:load', function() {

thenapking: I will double check this.

app/assets/javascripts/pdf_viewer/base.js
thenapking: I will also double check this.

thenapking: As I understand it the docs say this will require all files in app/assets/javascripts/pdf_viewer and below: https://guides.rubyonrails.org/asset_pipeline.html#manifest-files-and-directives

thenapking: I have inspected the page and on the PDF Viewer only the JS and CSS from Hypothesis is being loaded:
image
On the main page, the PDF Viewer JS is not loaded:
image

app/assets/javascripts/pdf_viewer/web/pdfjs-init.js
thenapking: Since turbolinks isn't being loaded, and neither is JQuery I don't think this is necessary at all.  I'd have to require both of those things additionally and I am not sure why I would want to.  It's also worth noting that this is kinda a cul-de-sac in the app - there are no links or other elements belonging to the app itself, only what is supplied by Hypothes.is.

PR #566, opened 2021-12-06
refactor reindex_available_works_job & reindex_model_jobs to accept ids for works to index

mankind:
We need to be able to pass the cnames of tenants to mint doi to each requeue of the job class. 
Secondly, **Tom W** wants us to accept a few ids to mint doi for to test it is working on demo sites, that way we don't have to loop through all the works in a work type  just to test that doi minting can be triggered by the Sidekiq job unless we are running it against production.

In the test calling this:

    expect { described_class.perform_now(work.class.to_s, account.cname, limit: 1, options: options) }.to change { work.doi_status_when_public }.to eq "findable"

 always  returns 
 
    expect X to have changed to eq "findable", but did not change 
 
Hence you will see that I used:

     described_class.perform_now(work.class.to_s, account.cname, limit: 1, options: options)
     work.reload
     expect(work.doi_status_when_public).to eq "findable"

### DIFF
--- a/app/jobs/hyku_addons/reindex_available_works_job.rb
+++ b/app/jobs/hyku_addons/reindex_available_works_job.rb
@@ -1,8 +1,16 @@
 # frozen_string_literal: true
+
+# call with
+# HykuAddons::ReindexAvailableWorksJob.perform_now(['repo.hyku.docker'], cname_doi_mint: ["repo.hyku.docker"] )
+#
+# HykuAddons::ReindexAvailableWorksJob.perform_now(['repo.hyku.docker'])
+
 module HykuAddons
   class ReindexAvailableWorksJob < ApplicationJob
     # cnames is an array
-    def perform(cnames)
+    def perform(cnames, cname_doi_mint: [])
+      @cname_doi_mint = cname_doi_mint
+
       cnames.each do |name|
         cname = cnames.delete(name)
         reindex(cname)
@@ -21,7 +29,8 @@ module HykuAddons
           next unless klass.count.positive?
 
           Rails.logger.debug "==== Queue ReindexModelJob for #{klass} ==="
-          HykuAddons::ReindexModelJob.perform_later(model_class, cname)
+          HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint }) if @cname_doi_mint.include?(cname)
+          HykuAddons::ReindexModelJob.perform_later(model_class, cname) unless @cname_doi_mint.include?(cname)
         end
       end
   end
--- a/app/jobs/hyku_addons/reindex_available_works_job.rb
+++ b/app/jobs/hyku_addons/reindex_available_works_job.rb
@@ -12,8 +12,7 @@ module HykuAddons
       @cname_doi_mint = cname_doi_mint
 
       cnames.each do |name|
-        cname = cnames.delete(name)
-        reindex(cname)
+        reindex(cnames.delete(name))
       end
     end
 
@@ -29,8 +28,7 @@ module HykuAddons
           next unless klass.count.positive?
 
           Rails.logger.debug "==== Queue ReindexModelJob for #{klass} ==="
-          HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint }) if @cname_doi_mint.include?(cname)
-          HykuAddons::ReindexModelJob.perform_later(model_class, cname) unless @cname_doi_mint.include?(cname)
+          HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint })
         end
       end
   end
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -1,43 +1,38 @@
 # frozen_string_literal: true
+
+# call with
+#  use_work_ids an array is just for testing a few ids
+# HykuAddons::ReindexModelJob.perform_now("GenericWork", "repo.hyku.docker", limit: 1, options: { cname_doi_mint: ["repo.hyku.docker"], use_work_ids: ["15b3c16b-8183-4d2a-86ef-d144176fa0c8"]} )
+#
+# HykuAddons::ReindexAvailableWorksJob.perform_later(["repo.hyku.docker"], options: { cname_doi_mint: ["repo.hyku.docker"]} )
+
 module HykuAddons
   class ReindexModelJob < ApplicationJob
-    queue_as :reindex
-
     rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused do |exception|
       Rails.logger.debug exception.inspect
       retry_job(wait: 5.minutes)
     end
 
-    def perform(klass, cname, limit: 25, page: 1, cname_doi_mint: [])
+    def perform(klass, cname, limit: 25, page: 1, options: {})
+      options ||= options || { cname_doi_mint: [], use_work_ids: [] }
       # for whatever in private methods reason without assigning it to instamce variable it throws undefined local variable
-      @cname_doi_mint = cname_doi_mint
+      @cname_doi_mint = options[:cname_doi_mint]
+      @use_work_ids = options[:use_work_ids]
       @cname = cname
+      @limit = limit
+      @page = page
+      @klass = klass
       AccountElevator.switch!(cname)
 
-      Rails.logger.debug "=== Starting to reindex #{klass} in #{cname} ==="
-
-      offset = (page - 1) * limit
-
-      # When the offset becomes too large, no records would be found
-      works =  klass.constantize.where("title_tesim:*").limit(limit).offset(offset)
-
-      # with calling to_a it always returns true, even when no records found
-      if works.to_a.any?
-        reindex_works(works)
-
-        new_page_count = page.to_i + 1
-
-        # Re-enqueue
-        ReindexModelJob.perform_later(klass, cname, page: new_page_count)
-      end
-
-      Rails.logger.debug "=== Completed reindex of #{klass} in #{cname} ==="
+      offset = (page - 1) * @limit
+      work_class = klass.constantize
+      fetch_work_using_ids(work_class) if @use_work_ids.present?
+      reindex_and_mint_work(work_class, offset) unless @use_work_ids.present?
     end
 
     private
 
       def reindex_works(works)
-        # works.each(&:update_index)
         works.each do |work|
           mint_doi(work) if @cname_doi_mint.present? && @cname_doi_mint.include?(@cname)
 
@@ -51,9 +46,37 @@ module HykuAddons
       def mint_doi(work)
         return if work.doi.present?
 
-        work.update(doi_status_when_public: "findable")
+        Rails.logger.debug "=== about to mint doi for #{work.title} ==== "
+        work.update(doi_status_when_public: "findable", visibility: "open")
         register_doi = Hyrax::DOI::DataCiteRegistrar.new.register!(object: work)
         work.update(doi: [register_doi.identifier])
+      end
+
+      def fetch_work_using_ids(klass)
+        @use_work_ids.map do |id|
+          work = klass.find(id)
+          Rails.logger.debug "=== updating index with #{id}for #{work&.title&.to_a&.first} ===="
+          mint_doi(work) if @cname_doi_mint.present? && @cname_doi_mint.include?(@cname)
+          work.save
+        end
+      end
+
+      # When the offset becomes too large, no records would be found
+      def reindex_and_mint_work(work_class, offset)
+        works = work_class.where("title_tesim:*").limit(@limit).offset(offset)
+        return unless works&.to_a&.any?
+
+        Rails.logger.debug "=== Starting to reindex #{@klass} in #{@cname} ==="
+
+        # with calling to_a it always returns true, even when no records found
+        reindex_works(works)
+
+        new_page_count = @page.to_i + 1
+
+        # Re-enqueue
+        ReindexModelJob.perform_later(@klass, @cname, @limit, page: new_page_count, cname_doi_mint: @cname_doi_mint)
+        sleep(0.5)
+        Rails.logger.debug "=== Completed reindex of #{@klass} in #{@cname} ==="
       end
   end
 end
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -32,20 +32,8 @@ module HykuAddons
 
     private
 
-      def reindex_works(works)
-        works.each do |work|
-          mint_doi(work) if @cname_doi_mint.present? && @cname_doi_mint.include?(@cname)
-
-          # We have temporarily replaced work.update_index with a  save to kill two birds with
-          # one stone, as in re_index and also update member_of_collections_ssim to store
-          # collection names instead of id caused by wrong bulkrax import
-          work.save
-        end
-      end
-
       def mint_doi(work)
         return if work.doi.present?
-
         Rails.logger.debug "=== about to mint doi for #{work.title} ==== "
         work.update(doi_status_when_public: "findable")
         register_doi = Hyrax::DOI::DataCiteRegistrar.new.register!(object: work)
@@ -61,10 +49,21 @@ module HykuAddons
         end
       end
 
+      def reindex_works(works)
+        works.each do |work|
+          mint_doi(work) if @cname_doi_mint.present? && @cname_doi_mint.include?(@cname)
+
+          # We have temporarily replaced work.update_index with a  save to kill two birds with
+          # one stone, as in re_index and also update member_of_collections_ssim to store
+          # collection names instead of id caused by wrong bulkrax import
+          work.save
+        end
+      end
+
       # When the offset becomes too large, no records would be found
       def reindex_and_mint_work(work_class, offset)
-        works = work_class.where("title_tesim:*").limit(@limit).offset(offset)
-        return unless works&.to_a&.any?
+        works = work_class.where("title_tesim:*").limit(@limit).offset(offset)&.to_a
+        return unless works&.any?
 
         Rails.logger.debug "=== Starting to reindex #{@klass} in #{@cname} ==="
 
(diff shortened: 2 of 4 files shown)


app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,38 @@
...
   class ReindexModelJob < ApplicationJob
-    queue_as :reindex
-
     rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused do |exception|
       Rails.logger.debug exception.inspect
       retry_job(wait: 5.minutes)
     end
 
-    def perform(klass, cname, limit: 25, page: 1, cname_doi_mint: [])
+    def perform(klass, cname, limit: 25, page: 1, options: {})
```
thenapking, line 16: There are no tests updated for this class.

app/jobs/hyku_addons/reindex_available_works_job.rb
```diff
@@ -21,7 +29,8 @@ def reindex(cname)
           next unless klass.count.positive?
 
           Rails.logger.debug "==== Queue ReindexModelJob for #{klass} ==="
-          HykuAddons::ReindexModelJob.perform_later(model_class, cname)
+          HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint }) if @cname_doi_mint.include?(cname)
```
prdanelli, line 32: I know that Hyrax/Hyku do this a lot, add extra conditions at the end of lines for `if/else`, but it woudl be clearer if we instead laid out a concise condition: 

```
options = @cname_doi_mint.include?(cname) ? { options: { cname_doi_mint: @cname_doi_mint } } : {}

HykuAddons::ReindexModelJob.perform_later(model_class, cname, options) 
```

app/jobs/hyku_addons/reindex_available_works_job.rb
```diff
@@ -1,8 +1,16 @@
...
+# call with
+# HykuAddons::ReindexAvailableWorksJob.perform_now(['repo.hyku.docker'], cname_doi_mint: ["repo.hyku.docker"] )
+#
+# HykuAddons::ReindexAvailableWorksJob.perform_now(['repo.hyku.docker'])
+
 module HykuAddons
   class ReindexAvailableWorksJob < ApplicationJob
     # cnames is an array
-    def perform(cnames)
+    def perform(cnames, cname_doi_mint: [])
```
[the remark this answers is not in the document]
mankind: Because like time we spoke last time, the test was testing perform and I will refactor to test the parameters passed in but even it that it does not give me any confidence that testing that the right parameters were passed means the job will function as expected.
This is why i have not updated the test.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,38 @@
...
-        new_page_count = page.to_i + 1
-
-        # Re-enqueue
-        ReindexModelJob.perform_later(klass, cname, page: new_page_count)
-      end
-
-      Rails.logger.debug "=== Completed reindex of #{klass} in #{cname} ==="
+      offset = (page - 1) * @limit
+      work_class = klass.constantize
+      fetch_work_using_ids(work_class) if @use_work_ids.present?
```
prdanelli, line 29: Can we use a more immediately apparent condition?

```
if @use_work_ids.present?
    fetch_work_using_ids(work_class) 
else
    reindex_and_mint_work(work_class, offset) 
end
```

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -51,9 +46,37 @@ def reindex_works(works)
...
+        Rails.logger.debug "=== Starting to reindex #{@klass} in #{@cname} ==="
+
+        # with calling to_a it always returns true, even when no records found
+        reindex_works(works)
+
+        new_page_count = @page.to_i + 1
+
+        # Re-enqueue
+        ReindexModelJob.perform_later(@klass, @cname, @limit, page: new_page_count, cname_doi_mint: @cname_doi_mint)
+        sleep(0.5)
```
prdanelli, line 78: I'm not sure we should be adding sleep's to the jobs, this seems like a smell to me.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: So what is your suggestion @thenapking

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -51,9 +46,37 @@ def reindex_works(works)
...
+
+        Rails.logger.debug "=== Starting to reindex #{@klass} in #{@cname} ==="
+
+        # with calling to_a it always returns true, even when no records found
+        reindex_works(works)
+
+        new_page_count = @page.to_i + 1
+
+        # Re-enqueue
+        ReindexModelJob.perform_later(@klass, @cname, @limit, page: new_page_count, cname_doi_mint: @cname_doi_mint)
```
prdanelli, line 77: You dont need to add  new variable for the updated page count, the code is more readable without it: 

```ReindexModelJob.perform_later(@klass, @cname, @limit, page: @page.to_i + 1, cname_doi_mint: @cname_doi_mint)```

**Review by prdanelli — CHANGES_REQUESTED** [review]

File: app/jobs/hyku_addons/reindex_available_works_job.rb (hunk 2 above)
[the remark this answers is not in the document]
Comment (mankind) [reply app/jobs/hyku_addons/reindex_available_works_job.rb:32]: What do you think about just: 

                HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint })
                
This is because @cname_doi_mint always returns a fault of empty array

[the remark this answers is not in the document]
Comment (prdanelli) [reply app/jobs/hyku_addons/reindex_available_works_job.rb:32]: I said something very similar @thenapking. but as the `options` hash is set to a blank hash by default, if could be a ternary. 

```options = @cname_doi_mint.include?(cname) ? { options: { cname_doi_mint: @cname_doi_mint } } : {}

HykuAddons::ReindexModelJob.perform_later(model_class, cname, options) ```

File: app/jobs/hyku_addons/reindex_available_works_job.rb [hunk 7]
```diff
@@ -1,8 +1,16 @@
...
+      @cname_doi_mint = cname_doi_mint
+
       cnames.each do |name|
```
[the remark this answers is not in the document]
mankind: Ok , noted will make the change, thanks

app/jobs/hyku_addons/reindex_model_job.rb
mankind: I am happy to remove but what makes it a smell

thenapking: Especially not after asking a job to `perform_later`.

prdanelli: Just adding in sleep within a background job. A job should ideally just perform a single task, and if there are more tasks, each should be delegated. So having a sleep here, would imply that the job is doing a lot of work that needs to be broken up to prevent the site's performance from being hindered.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: I can make that change though i don't find your change easier to process if glancing quickly through the code compared to the current code.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: I thought you loved saving lines, was done with you in mind.

I will change it though/

app/jobs/hyku_addons/reindex_available_works_job.rb
mankind: I think this will do:
               HykuAddons::ReindexModelJob.perform_later(model_class, cname, options: { cname_doi_mint: @cname_doi_mint })

@cname_doi_mint is always an empty array, so I am not sure we need the check here:
    options = @cname_doi_mint.include?(cname) ? { options: { cname_doi_mint: @cname_doi_mint } } : {}

app/jobs/hyku_addons/reindex_model_job.rb
prdanelli: hahaha. I do like saving lines, but clarity comes first  :D

I think its just this pattern that Hyrax uses, with these kind of end of line conditions for if/else... IMO they are not easier to read.

app/jobs/hyku_addons/reindex_model_job.rb
thenapking: This class calls save on some works, presumably in order to trigger reindexing them.  We could at the very least check that the correct works have been updated.  We can also check that the work's attribute `doi_status_when_public` gets set to `findable` if the work has a doi. 
As there are now two options that can be passed in to this job, I would expect there to be at least two test scenarios covering the use of each of the options.

app/jobs/hyku_addons/reindex_available_works_job.rb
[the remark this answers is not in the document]
thenapking: Insofar as all this job does is spawn other jobs, the only thing I would expect to see in an updated test is that the spawned jobs have the new parameter passed through.  That's not the case with `ReindexModelJob`.

app/jobs/hyku_addons/reindex_available_works_job.rb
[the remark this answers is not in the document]
thenapking: Yep, this is clearer to read as well I think.

app/jobs/hyku_addons/reindex_model_job.rb
thenapking: I agree.  You enqueue a job to be processed later, then wait half a second, then finish writing to the log.  What /does/ happen in that half a second, and why is it so important to sleep before writing your final comment to the log? There should be no reason to wait to write to the log, and it indicates a fragility or failure elsewhere.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: I am making the change but @prdanelli going with your desk analogy, a method with 2 to 3 local variables is like a desk with 2 or 3 books on it being called cluttered. Sounds more like a preference of having nothing on the desk and not a case of cluttering.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,55 @@
 # frozen_string_literal: true
+
+# call with
+#  use_work_ids an array is just for testing a few ids
```
prdanelli, line 4: Could you tidy this comment up?

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,55 @@
...
-
-      # When the offset becomes too large, no records would be found
-      works =  klass.constantize.where("title_tesim:*").limit(limit).offset(offset)
-
-      # with calling to_a it always returns true, even when no records found
-      if works.to_a.any?
-        reindex_works(works)
+      offset = (page - 1) * @limit
+      work_class = klass.constantize
+      fetch_work_using_ids(work_class) if @use_work_ids.present?
```
prdanelli, line 29: I realise Rubocop gave you a warning about method length, but it could be resolved by removing the `offset` variable: 
```
work_class = klass.constantize
if @use_work_ids.present?
      fetch_work_using_ids(work_class) 
else
     # The offset variable is documented by the method signature
      reindex_and_mint_work(work_class, (page - 1) * @limit) 
end
```

It's better to add  `#rubocop:disable Metrics/MethodLength` and have a normal condition than two end of line conditions for `if` and `unless`.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,55 @@
...
+      reindex_and_mint_work(work_class, offset) unless @use_work_ids.present?
+    end
 
-        new_page_count = page.to_i + 1
+    private
 
-        # Re-enqueue
-        ReindexModelJob.perform_later(klass, cname, page: new_page_count)
+      def mint_doi(work)
+        return if work.doi.present?
```
prdanelli, line 36: Could you add a break return under your guard statements? It will make the code easier to read.

prdanelli (commented):
I left a couple of comments. I think if we can tidy them up just a little more then its perfect.

spec/jobs/reindex_model_job_spec.rb
```diff
@@ -14,11 +29,10 @@
...
   end
 
-  it "#perform_later can enqueue model for reindex later" do
-    expect { described_class.perform_later(work.class.to_s, account.cname) }.to have_enqueued_job(described_class).exactly(:once)
-  end
-
-  it "#perform_now will enqueue model for reindex" do
-    expect { described_class.perform_now(work.class.to_s, account.cname) }.to have_enqueued_job(described_class).exactly(:once)
+  it "sets doi_status_when_public to findable" do
+    expect(work.doi).to be_empty
```
thenapking, line 33: We don't need to test this, it is defined as empty on line 5.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: ok i will add.

spec/jobs/reindex_model_job_spec.rb
```diff
@@ -14,11 +29,10 @@
...
+    described_class.perform_now(work.class.to_s, account.cname, limit: 1, options: options)
+    work.reload
+    expect(work.doi_status_when_public).to eq "findable"
```
thenapking, line 36: You can combine this into one line, as it is standard in job tests to have to reload the object changed by the job.
````
expect(work.reload.doi_status_when_public).to eq "findable"
````

File: app/jobs/hyku_addons/reindex_model_job.rb (hunk 9 above)
Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:29]: @prdanelli one of the point of the Rubocops is to help to provide pointers to things to refactor. So i don't really think we should always defer to adding **#rubocop:disable Metrics/MethodLength** rather than actually refactor the code. Otherwise what is point of the rubocop.

I am not convinced of the benefits of this change vs what is there?

File: spec/jobs/reindex_model_job_spec.rb [hunk 13]
```diff
@@ -1,11 +1,26 @@
 # frozen_string_literal: true
 
 RSpec.describe HykuAddons::ReindexModelJob do
-  let(:account) { create(:account) }
-  let(:work) { create(:work) }
+  let(:account) { create(:account, cname: cname) }
+  let(:work) { create(:work, doi: []) }
+  let(:second_work) { create(:work, title: ["another work"], doi: []) }
```
thenapking, line 6: This isn't referenced anywhere in the test so we don't need to create it.

spec/jobs/reindex_model_job_spec.rb
```diff
@@ -1,11 +1,26 @@
...
+  let(:prefix) { "10.1234" }
+  let(:cname) { "123abc" }
+  let(:response_body) { File.read(HykuAddons::Engine.root.join("spec", "fixtures", "doi", "mint_doi_return_body.json")) }
+  let(:options) { { cname_doi_mint: [account.cname], use_work_ids: [work.id] } }
 
   before do
-    allow(work.class).to receive(:reindex_everything)
+    GenericWork.all.map(&:destroy)
```
thenapking, line 13: Why do we have to destroy all GenericWorks? Why have any objects been created for this test that we don't want created?

app/jobs/hyku_addons/reindex_model_job.rb
prdanelli: Thank you buddy.

spec/jobs/reindex_model_job_spec.rb
prdanelli: FYI, `GenericWork.destroy_all` would be a slightly nicer way of writing this.

app/jobs/hyku_addons/reindex_model_job.rb
thenapking: Thanks for adding the two test classes! I think this vastly improves the change.

spec/jobs/reindex_model_job_spec.rb
prdanelli: I just checked to make sure ActiveFedora implemented it: 

```
2.7.4 :001 > GenericWork.methods.include?(:destroy_all)
 => true
```

mankind: @thenapking that ought to be taken out, I was checking something locally and just forgot to take off.

spec/jobs/reindex_model_job_spec.rb
mankind: @thenapking if you mean **second_work** it  was used in the test with issues that i spoke about in the stand up  i removed, so it ought to be off and will be removed.

spec/jobs/reindex_model_job_spec.rb
mankind: @thenapking  despite I discovered sometimes the work in the test I removed that was using **second_work** still had doi set, so this was added just for me to be double sure the doi is empty or it fails.

mankind: @thenapking  despite that, I discovered that sometimes for the work in the test I removed that was using **second_work** the doi was set, when ought not to be so this was added just for me to be double sure tat the point of triggering the job that he doi was indeed empty otherwise the test  should fail.

I am happy to remove though.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -1,43 +1,58 @@
...
-
     rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused do |exception|
       Rails.logger.debug exception.inspect
       retry_job(wait: 5.minutes)
     end
 
-    def perform(klass, cname, limit: 25, page: 1, cname_doi_mint: [])
+    # rubocop:disable Metrics/MethodLength
+    def perform(klass, cname, limit: 25, page: 1, options: {})
+      options ||= options || { cname_doi_mint: [], use_work_ids: [] }
```
prdanelli, line 16: This can just be `options ||= { cname_doi_mint: [], use_work_ids: [] }`

app/jobs/hyku_addons/reindex_model_job.rb
thenapking: If this is something you feel strongly about we can add EmptyLineAfterGuardClause to rubocop.yml and then that will apply this automatically.

prdanelli (approved):

PR #567, opened 2021-12-07
Tweak the task to only try and regenerate existing pdf thumbnails

prdanelli:


mankind:
Can this not be one line:

            next if work.thumbnail.blank?
            next unless work.thumbnail.mime_type.match?(/^(.)+\/pdf$/)
            
 to this using either **not** or **!**
 
          next if work.thumbnail.blank? && not work.thumbnail.mime_type.match?(/^(.)+\/pdf$/)

mankind (approved):
Thanks :+1:

PR #568, opened 2021-12-07
REPO 1520 Refactor test suite inc ActiveJob configuration in

thenapking:
Currently the test suite takes more than 1 hour to run on a developer's machine which makes local testing practically impossible.  By analysing the test suite to remove unnecessary database calls, adding bootsnap, and disabling slow running feature tests from a standard run, I can now run the local test suite in **<10 minutes**.  The full suite will still run on Circle CI.

### Work Completed
- [x] Use `build_stubbed` instead of `create` where possible to increase speed
- [x] Ensure tests only invoke jobs required to test the features they are involved in
- [x] Exclude tests labelled `:slow` and `:customer` from standard test runs
- [x] Remove two empty tests in `csv_entry_validation_service_spec.rb` and `redlands_csv_entry_validation_service_spec.rb`
- [x] Create an rspec binstub; update README to show preferable way of running tests.
- [x] Ensure tests have a named type
- [x] Use bootsnap in development to speed up test suite
- [x] Add profiling tools (TestProf, RubyProf, StackProf)
- [x] Move files defining shared contexts to spec/support, instead of mixing them up with helper specs
- [x] Cache rubocop


spec/features/bulkrax_export_spec.rb
```diff
@@ -2,34 +2,37 @@
...
-  let(:work) { create(:fully_described_work, user: depositor.email) }
-  let(:account) { create(:account) }
 
-  before do
-    # Make sure default admin set exists
-    AdminSet.find_or_create_default_admin_set_id
-  end
+  let(:account) { build_stubbed(:account) }
 
   after do
```
prdanelli, line 24: Whilst we're doing this big edit, we might as well clean up stuff thats obviously never going to be done. :)

spec/features/bulkrax_export_spec.rb
```diff
@@ -63,15 +69,7 @@
   end
 
   describe "round-tripping" do
-    let(:exporter) do
-      create(:bulkrax_exporter_worktype,
```
prdanelli, line 67: Was this not required?

spec/features/create_anschutz_work_spec.rb
```diff
@@ -3,7 +3,7 @@
 include Warden::Test::Helpers
 
 # NOTE: If you generated more than one work, you have to set "js: true"
-RSpec.feature "Create a AnschutzWork" do
+RSpec.feature "Create a AnschutzWork", customer: true do
```
prdanelli, line 6: Is there a list of the keys that are available? There is a `slow: true` above and `customer: true` here. Whats the difference between them? Could we have just one so that we don't add extra complexity?

spec/rails_helper.rb
```diff
@@ -140,6 +140,11 @@
     clear_enqueued_jobs
   end
 
+  unless ENV['CI']
```
prdanelli, line 143: Is there a way to easily override this locally so we can run all of the specs as required?

prdanelli (dismissed):
I've added some comment, but overall this is amazing.

spec/rails_helper.rb
thenapking: Yes, will add instructions to the readme.

spec/features/create_anschutz_work_spec.rb
thenapking: There is a difference, but perhaps we could find a different name for the `:customer` tag which makes it more obvious.  The `:customer` tag is meant to apply specifically to feature tests which are all named `create_customer_name_work_type`.  So the generic tests, which don't have usage specific to a particular customer are still included in the test suite.  Perhaps it would be better if I added a tag per customer, so if you were altering things for a particular customer you could easily run those tests from the command line?  Slow tests are ones that will always be slow, that we need to run but you just don't want to run every time unless you are working on that part of the app.  You might run the full suite once including these tests before running CircleCI (which will be even slower).

spec/features/bulkrax_export_spec.rb
thenapking: There were several definitions for this, which varied only in their export_source, so I put the definition up top, and had individual test set their export source to make it more dry.

spec/features/create_anschutz_work_spec.rb
prdanelli: I think that might be overkill. We can already run a specific test or series of tests by specifying the specs file location and line number. 

I like the `slow: true`, as it could be applied to anything else, even the custom ones.

thenapking: I've checked this with @BertZZ and he thought customer name would be a good way to go.  It makes the tag clear.  Also I doubt you'd ever want to run all customers at once (outside of CircleCI) and you can still pick several to run at once.

prdanelli: I think the issue is that we have two tags to do the same thing. The customer specs are the slow specs, so tagging them differently will only lead to the tags used interchangably.

thenapking: Now only using `:slow` tag.

prdanelli (approved):

PR #570, opened 2021-12-09
Speed up the getter methods for ISNI and ROR

prdanelli:
The previous method used a `find` method within a loop that would iterate over each of the items for each of the loop iterations, so took about 1.5 seconds for each time the value was requested. 

```
          user       system     total      real
legacy    1.577007   0.016105   1.593112   (1.593188)
improved  0.000095   0.000000   0.000095   (0.000096)
```

### DIFF
--- a/app/services/hyku_addons/current_he_institution_service.rb
+++ b/app/services/hyku_addons/current_he_institution_service.rb
@@ -16,10 +16,10 @@ module HykuAddons
 
     protected
 
-    # This method accesses a private method on the authority which returns the raw yaml data as an array of hashes
-    # without removing elements. This means we can increase the speed with which we get different sets of keys
-    def quick_active_elements
-      authority.send(:terms).select { |a| a["active"] == true }
-    end
+      # This method accesses a private method on the authority which returns the raw yaml data as an array of hashes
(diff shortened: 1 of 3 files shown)


thenapking (commented):
This looks great.  Only one minor thing:  I think we are missing a test for when either `authority.send(:terms)` is `nil` or all the elements are not active.

thenapking:
> > This looks great. Only one minor thing: I think we are missing a test for when either `authority.send(:terms)` is `nil` or all the elements are not active.
> 
> Thanks @thenapking. In this case I dont think its an issue because the yaml always has a backup in the live config.
> 
> One part of the spec i wasn't sure about was this: https://github.com/ubiquitypress/hyku_addons/pull/570/files#diff-c46e9bad23c23977f14d37eb8e21edc915d8c30786eb1ef6142f1bb3907b1f98R15
> 
> Is this the best way of doing this?

I'm not sure if there is a better way of loading that test fixture.  How is that yaml normally loaded?  I guess regarding the nil edge cases I was more thinking about testing whether the code was written to handle those nils, irrespective of what data might get loaded.

thenapking (approved):
LGTM!

PR #571, opened 2021-12-10
reindex-job should not mint doi when the work is private or under review

mankind:
This refactor prevents the ReindexModelJob from minting doi when a work is private or when the work is under review.

### DIFF
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -35,12 +35,17 @@ module HykuAddons
     private
 
       def mint_doi(work)
-        return if work.doi.present?
+        return if work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
 
         Rails.logger.debug "=== about to mint doi for #{work.title} ==== "
+
         work.update(doi_status_when_public: "findable")
         register_doi = Hyrax::DOI::DataCiteRegistrar.new.register!(object: work)
         work.update(doi: [register_doi.identifier])
+      end
+
+      def workflow_state(work)
+        @_workflow_state ||= Sipity::Entity.find_by(proxy_for_global_id: "gid://hyku/#{work.class}/#{work.id}")
       end
 
       def fetch_work_using_ids(klass)
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -35,13 +35,17 @@ module HykuAddons
     private
 
       def mint_doi(work)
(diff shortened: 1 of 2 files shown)


spec/jobs/reindex_model_job_spec.rb
```diff
@@ -24,11 +33,39 @@
...
+      described_class.perform_now(work.class.to_s, account.cname, limit: 1, options: options)
+      expect(work.reload.doi_status_when_public).to eq "findable"
+    end
   end
 
-  it "sets doi_status_when_public to findable" do
-    described_class.perform_now(work.class.to_s, account.cname, limit: 1, options: options)
-    expect(work.reload.doi_status_when_public).to eq "findable"
+  context "skip doi minting" do
+    let(:sipity_entity) { instance_double(Sipity::Entity, proxy_for_global_id: "gid://hyku/#{pending_review_work.class}/#{pending_review_work.id}", workflow_id: workflow.id, workflow_state: sipity_workflow_state) }
```
prdanelli, line 54: Could we shorten these longer lines, they're getting kind of gnarly, this one and L44. 

```
let(:gid) { "gid://hyku/#{pending_review_work.class}/#{pending_review_work.id}" }
let(:sipity_entity) { instance_double(Sipity::Entity, proxy_for_global_id: gid, workflow_id: workflow.id, workflow_state: sipity_workflow_state) }
```

mankind: Thanks I will :+1:

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -35,14 +35,19 @@ def perform(klass, cname, limit: 25, page: 1, options: {})
     private
 
       def mint_doi(work)
-        return if work.doi.present?
+        return if work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
```
prdanelli, line 38: I find that when the guard clause starts to have more than one or two clauses to it, then it is better to add it to its own method. With this example there are positive and negative checks too. I would use something like this: 

```
return unless can_mint_for?(work)

def can_mint_for?(work)
    work.doi.blank? || work.visibility == "open" || workflow_state(work)&.workflow_state_name == "deposited"
end
```

mankind: Thanks will update it

mankind: I made the change to use unless with these conditions:

    work.doi.blank? || work.visibility == "open" || workflow_state(work)&.workflow_state_name == "deposited"
    
but the test failed so I reverted it back to what it was  but extracted it into a method like you did.
I thought i should let you know

prdanelli (approved):

PR #575, opened 2021-12-13
Collection banner in api

BertZZ:


spec/requests/api_v1_collection_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      date_created: [date_created],
+      collection_type_gid: collection_type_gid,
+      description: [description],
+      title: ["Test Title"],
+      visibility: "open"
+    }
+  end
+
+  let(:collection) { Collection.create(attributes) }
+  let(:collection_branding) { CollectionBrandingInfo.new(collection_id: collection.id, role: "banner", filename: "nypl-hydra-of-lerna.jpg").save("spec/internal_test_hyku/spec/fixtures/images/nypl-hydra-of-lerna.jpg", false) }
```
mankind, line 24: Will this file **fixtures/images/nypl-hydra-of-lerna.jpg"** in hyrax always be there.
Why did you not just give a string to local_path and then check the collection_banner_url matches it, so everything is with our control. What is the problem witrh passing it a local_path with a file path as string?

On calling new and then save rather than using either a factory or an instance double, we can keep that if you want

BertZZ: We use that file in other tests, its kind of like our test file. Giving it a fake path results in a file not found error when doing it like this. I think you make a good point about everything being in our controll. Give me a few minutes to try a refactor

mankind: Thanks @BertZZ

BertZZ:
I tried to get both tests in separate it blocks but kept getting ` #<ClassDouble(CollectionBrandingInfo) (anonymous)> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.
`

So I might try again tomorrow but I think this test is pretty good

mankind:
> I tried to get both tests in separate it blocks but kept getting `#<ClassDouble(CollectionBrandingInfo) (anonymous)> was originally created in one example but has leaked into another example and can no longer be used. rspec-mocks' doubles are designed to only last for one example, and you need to create a new one in each example you wish to use it for.`
> 
> So I might try again tomorrow but I think this test is pretty good

I think so too :+1:

mankind (approved):
👍

PR #578, opened 2021-12-14
ensure job requeues have the correct paramters

mankind:
The ReindexModelJob re-queues itself but was not doing so with the right parameters.  This adds a test to catch such issues next time.


thenapking (approved):
LGTM.

mankind:
Thank you @thenapking

PR #579, opened 2021-12-14
Ensure login not required

thenapking:
Feedback from QA showed that the display logic was wrong.
The PDF Controller should not check a user is logged in because this feature is available to all internet users.
Additionally, the link was not showing up in quite the correct circumstances.  In some cases the link was apparent when the feature flipper was not enabled.


app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if can?(:edit, file_set.id) || can?(:destroy, file_set.id) || (Flipflop.enabled?(:annotation) && can?(:download, file_set.id)) %>
+<% if can?(:edit, file_set.id) || can?(:destroy, file_set.id) || Flipflop.enabled?(:annotation) && file_set.pdf? && can?(:download, file_set.id) %>
```
mankind, line 1: Just checking out this condition

    Flipflop.enabled?(:annotation) && file_set.pdf? && can?(:download, file_set.id)
      
 Should   **can?(:download, file_set.id)**  not be something you can do even if the **file_set** is not a pdf provided it is not private.

prdanelli, line 1: Just a little thing, but could you add brackets around the and conditions? Just to make it clear what is connected to what?

```
can?(:edit, file_set.id) || can?(:destroy, file_set.id) || (Flipflop.enabled?(:annotation) && file_set.pdf? && can?(:download, file_set.id))
```

prdanelli (dismissed):

app/views/hyrax/file_sets/_actions.html.erb
prdanelli: I just noticed that you removed the brackets, but that does mean from reading the line i'm not sure exactly what conditions are connected.

thenapking: Ah yes.  You are correct.  There's a lot of conditions on this page and it's easy to get lost in them.

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -39,18 +39,26 @@
...
     </ul>
   </div>
+<% elsif Flipflop.enabled?(:annotation) && file_set.pdf? %>
+  <%= link_to t('.read_and_annotate'), 
+  hyku_addons.pdf_viewer_path(hyrax.download_url(file_set.id).split('/')&.last&.split('?')&.first),  
+  title: 'Annotate', 
+  id: "file_read", 
+  target: "_blank",
+  class: 'btn btn-default btn-sm' %>
 <% elsif can?(:download, file_set.id) %>
```
mankind, line 57: can?(:download, file_set.id) in line 57 is a repeat of line 32, I am trying to understand why we need hem in both places?

mankind: Secondly if it turns out that we need it in both places, should it move into  a partial for sharing?

prdanelli: @mankind keep in mind that using partials is closer than keeping everything in the same place.:)

app/views/hyrax/file_sets/_actions.html.erb
thenapking: That was just a mistake on my part during refactoring.  I'll put the brackets back.

app/views/hyrax/file_sets/_actions.html.erb
thenapking: One of the functions of this partial is to decide whether to render a dropdown list or a single button.  If there are multiple conditions you fulfil then we need a dropdown.  There are only two circumstances where you can only fulfil one: when you can only annotate or only download.  Once we've decided that a dropdown is needed we still need to check which options to fill it with, which is why there are individual checks matching each of the conditions.  I will refactor this class to make this more clear, as I have tried to get this right several times and it still remains obscure.

mankind: Ok thanks @thenapking  for the clarification.

thenapking: Addressed below.

app/views/hyrax/file_sets/_actions.html.erb
thenapking: Addressed below.

thenapking: Addressed below.

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,57 +1,6 @@
...
-    </ul>
-  </div>
-<% elsif can?(:download, file_set.id) %>
-  <%= link_to t('.download'),
-    hyrax.download_path(file_set),
-    title: t('.download_title', file_set: file_set),
-    target: "_blank",
-    id: "file_download",
-    data: { label: file_set.id } %>
+<% view_condition = [can?(:edit, file_set.id), can?(:destroy, file_set.id), can?(:download, file_set.id), (Flipflop.enabled?(:annotation) && file_set.pdf?)] %>
```
thenapking, line 1: Is there somewhere better I can define this variable? I see definitions like this scattered throughout the views, which is not ideal, but maybe necessary.  Let me know if I can improve this.

mankind: will a helper make sense?

mankind (approved):

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,57 +1,6 @@
...
+<% if can?(:edit, file_set.id) || (view_condition.any? && !view_condition.one?)  %>
+  <%= render partial: 'hyrax/file_sets/actions_list', locals: {file_set: file_set} %>
+<% elsif view_condition.one? %>
+  <%= render partial: 'hyrax/file_sets/actions_button', locals: {file_set: file_set} %>
 <% end %>
```
prdanelli, line 6: Im not sure about how this is written as im struggling a little how the condition will compute. So, any of the `view_condition`s are true but not just one of them? And the array of conitions contains `can?(:edit, file_set.id)` as well as the actual condition, which i assume is why you have the `!.one?`. 

I think its a smell that the array of conditions is 162 characters long. I would suggest its best to break up the different conditions into more manageable chunks and then the if condition will be far easier to comprehend.

```ruby
<% args = { file_set: file_set } %>
<% can_annotate = Flipflop.enabled?(:annotation) && file_set.pdf? %>
<% can_edit = can?(:edit, file_set.id) %>
<% can_maintain = can_edit && (can?(:destroy, file_set.id) || can?(:download, file_set.id)) %>

<% if can_maintain && can_annonate  %>
  <%= render partial: "hyrax/file_sets/actions_list", locals: args %>
<% elsif can_edit %>
  <%= render partial: "hyrax/file_sets/actions_button", locals: args %>
<% end %>
```

I'm not 100% sure if the `can_maintain` conditions match yours, its just an example of how i'd tackle a condition like this. I also wish we could use `?` in variable names as they'd be so much nicer.

app/views/hyrax/file_sets/_actions_button.html.erb
```diff
@@ -0,0 +1,27 @@
...
+  <%= link_to t('.read_and_annotate'), 
+    hyku_addons.pdf_viewer_path(hyrax.download_url(file_set.id).split('/')&.last&.split('?')&.first),  
+    title: 'Annotate', 
+    id: "file_read", 
+    target: "_blank",
+    class: 'btn btn-default btn-sm' %>
+<% elsif can?(:download, file_set.id) %>
+  <%= link_to t('.download'),
+    hyrax.download_path(file_set),
+    title: t('.download_title', file_set: file_set),
```
prdanelli, line 22: You've got a mix of single/double quotes in here :)

prdanelli (commented):
I have added a couple of comments.

app/views/hyrax/file_sets/_actions.html.erb
mankind: I think everything should be moved to a helper and we can just called **can_maintain && can_annonate ** which will be both helpers or what ever new name they are called.
Or better still have just one helper to call in the view.

prdanelli: Helpers would require the fileset to be passed and also have access to the current request scope. I'm not sure we need helpers for a scenario where its only done once. If we needed to do this in a couple of places, then it would be best in a helper, but the `can?` method is already a helper.

PR #581, opened 2021-12-16
Add solr doi to see if it is read by work

prdanelli:


app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -33,14 +33,20 @@ json.cache! [@account, :works, work.id, work.solr_document[:_version_], work.mem
   json.date_submitted work.date_uploaded
   json.degree work.try(:solr_document)&.to_h&.dig('degree_tesim')
   json.dewey work.try(:solr_document)&.to_h&.dig('dewey_tesim')
+
   json.doi work.try(:doi)
+  if (doi = work.try(:solr_document)&.to_h&.dig('doi_ssi')).present?
+    json.solr_doi "https://doi.org/#{doi}"
```
BertZZ, line 39: This is changing the name of the field on the front end so can we keep it as `json.doi` to avoid any extra work for Rees

BertZZ (changes requested):

app/views/hyku/api/v1/work/_work.json.jbuilder
BertZZ, line 39: Why are we adding a new field?

BertZZ (approved):

PR #585, opened 2021-12-17
Move DOI tab into main description

thenapking:
This PR adds a feature flipper to control the placement of the DOI fields in the work form.

### Work completed
Create `HykuAddons::DOIFormBehavior` to override gem's behaviour
Update each work type to require above instead
Add method to Hyrax::DOI::WorkFormHelper to stop tab being displayed
Bring in forms and update layout
Update DOI Autofill test and DOI Minting


app/controllers/concerns/hyku_addons/doi_form_behavior.rb
```diff
@@ -0,0 +1,20 @@
...
+    extend ActiveSupport::Concern
+
+    included do
+      self.terms += [:doi]
+
+      delegate :doi, to: :model
+    end
+
+    def primary_terms
+      !Flipflop.doi_tab? && Flipflop.doi_minting? ? super << :doi : super
```
prdanelli, line 13: Instead of a ternary, could this be simplified a little by using a guard statement?

```ruby
def primary_terms
    return super if Flipflop.doi_tab? && !Flipflop.doi_minting?

    super << :doi
end
```

I'd usually argue that ternary conditions are good for `this or that`, rather than adding logic with side effects.

app/helpers/hyrax/doi/work_form_helper.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+module Hyrax
+  module DOI
+    module WorkFormHelper
+      def form_tabs_for(form:)
+        if Flipflop.doi_tab? && form.model_class.ancestors.include?(Hyrax::DOI::DOIBehavior)
```
prdanelli, line 6: This is the only time you're not negating the method; would it be better to rename the feature to be `inline_doi` and not have to negative it through out the PR?

prdanelli (commented):
Looks great. Just a couple of questions.

app/controllers/concerns/hyku_addons/doi_form_behavior.rb
thenapking: Absolutely

app/helpers/hyrax/doi/work_form_helper.rb
thenapking: Ah, yes. That's a good idea.  I could not think of an opposite name for "DOI tab" that wouldn't make it more confusing. But your suggestion will do it!

thenapking: Addressed below

lib/hyku_addons/engine.rb
```diff
@@ -495,6 +495,10 @@ def self.dynamically_include_mixins
       Image.include HykuAddons::ImageOverrides
       GenericWork.include ::Hyrax::BasicMetadata
       Hyrax::WorkIndexer.include HykuAddons::WorkIndexerBehavior
+
+      # HykuAddons::DOIFormBehavior must be prepended before WorkForm overrides
+      Hyrax::DOI::DOIFormBehavior.prepend HykuAddons::DOIFormBehavior
```
thenapking, line 500: Thanks for your helpful feedback @mankind this morning.  I tried `prepend`, which I hadn't before, as that was the correct way to override this class.  Still didn't work, but eventually I realised this needed to be prepended before the WorkForm overrides.

app/controllers/concerns/hyku_addons/doi_form_behavior.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  module DOIFormBehavior
+    def primary_terms
+      return super if !Flipflop.inline_doi? && !Flipflop.doi_minting?
```
prdanelli, line 6: Should we use `unless` here?

prdanelli (approved):
I left a single comment abot `if/unless` so if you decide to change that it would be good, but its not super important.

mankind (approved):
Thanks @thenapking  for making the changes :+1:   LGTM

prdanelli (dismissed):

mankind (approved):
:+1:

PR #586, opened 2021-12-17
Refactor action list

thenapking:
Following comments on https://github.com/ubiquitypress/hyku_addons/pull/579, this comment refactors the actions list to make the logic a little more clear.


app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,6 +1,9 @@
...
-<% if can?(:edit, file_set.id) || (view_condition.any? && !view_condition.one?)  %>
+<% can_annotate = Flipflop.enabled?(:annotation) && file_set.pdf? %>
+<% user_actions_count = [can?(:edit, file_set.id), can?(:destroy, file_set.id), can?(:download, file_set.id), can_annotate].count(true) %>
+<% list_item_count = can?(:edit, file_set.id) ? user_actions_count + 1 : user_actions_count  %>
+
+<% if list_item_count > 1  %>
   <%= render partial: 'hyrax/file_sets/actions_list', locals: {file_set: file_set} %>
-<% elsif view_condition.one? %>
+<% elsif list_item_count == 1 %>
   <%= render partial: 'hyrax/file_sets/actions_button', locals: {file_set: file_set} %>
```
prdanelli, line 8: Can we replace the hash duplication with something like: 

```
<% locals = { locals: { file_set: file_set } }  %>
# ...
<%= render partial: "hyrax/file_sets/actions_button", locals %>
```
Could you also add spaces inside the hash and switch the single to double quotes - just for consistency with the rest of the app.

app/views/hyrax/file_sets/_actions.html.erb
```diff
@@ -1,6 +1,9 @@
-<% view_condition = [can?(:edit, file_set.id), can?(:destroy, file_set.id), can?(:download, file_set.id), (Flipflop.enabled?(:annotation) && file_set.pdf?)] %>
-<% if can?(:edit, file_set.id) || (view_condition.any? && !view_condition.one?)  %>
+<% can_annotate = Flipflop.enabled?(:annotation) && file_set.pdf? %>
+<% user_actions_count = [can?(:edit, file_set.id), can?(:destroy, file_set.id), can?(:download, file_set.id), can_annotate].count(true) %>
+<% list_item_count = can?(:edit, file_set.id) ? user_actions_count + 1 : user_actions_count  %>
+
+<% if list_item_count > 1  %>
   <%= render partial: 'hyrax/file_sets/actions_list', locals: {file_set: file_set} %>
-<% elsif view_condition.one? %>
+<% elsif list_item_count == 1 %>
```
prdanelli, line 7: `list_item_count` doesn't really explain what this variable does. I'm not totally convincing by using counts for this, and where there is one or more than one. Wouldn't it be better to use booleans that are computed before? I'm just struggling to keep it in my head as I read through, so that strikes be as a little too complicated.

prdanelli (commented):
I left a couple of comments.

app/views/hyrax/file_sets/_actions.html.erb
thenapking: Added comment and renamed variables as per discussion.

prdanelli (approved):

PR #588, opened 2021-12-20
Hotfix/rescue blacklight error

prdanelli:
Attempt to prevent "The value for :adapter was not found in the blacklight.yml config" errors by rescuing the error and setting the value of the `adapter`. This will likely still see errors, as to set the value, we need to see an error, in which case maybe we can try manually setting the values in the PR in a controller before action.

### DIFF
--- a/app/controllers/concerns/hyku_addons/application_controller_override.rb
+++ b/app/controllers/concerns/hyku_addons/application_controller_override.rb
@@ -5,7 +5,21 @@ module HykuAddons
   module ApplicationControllerOverride
     extend ActiveSupport::Concern
 
+    included do
+      rescue_from RuntimeError, with: :blacklight_adapter_error
+    end
+
     private
+
+      # Try and resue the blacklight errors we have been seeing. However, this will likely either work, or move the
+      # error elsewhere within Sentry, so this should be considered a temporary fix at best.
+      def blacklight_adapter_error(error)
+        raise error unless error.message == "The value for :adapter was not found in the blacklight.yml config"
+
+        # I'm not 100% sure which to set, so I'm setting both
+        Blacklight.connection_config[:adapter] = "solr"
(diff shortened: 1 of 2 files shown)


BertZZ (approved):

app/controllers/concerns/hyku_addons/application_controller_override.rb
```diff
@@ -5,8 +5,22 @@ module HykuAddons
...
+
     private
 
+      # Try and resue the blacklight errors we have been seeing. However, this will likely either work, or move the
+      # error elsewhere within Sentry, so this should be considered a temporary fix at best.
+      def blacklight_adapter_error(error)
+        raise error unless error.message == "The value for :adapter was not found in the blacklight.yml config"
+
+        # I'm not 100% sure which to set, so I'm setting both
+        Blacklight.connection_config[:adapter] = "solr"
```
mankind, line 20: I have nothing against just setting the adapter, why not set the whole configuration object:
   Blacklight.connection_config

mankind: And  in addition, we will still keep the other setting that sets the repository class

prdanelli: I think that its probably best to take small steps when making temp fixes like this, and avoid overriding more than we absolutely need to. If we find that we fix this issue and start to see another error relating to the `url`, then at that point we have a good reason to also set that value too. 

As an aside, Im not sure the SolrService instance options is where to put this stuff, as i'm not seeing those keys in the hash. 

```
2.7.4 :002 > AccountElevator.switch! Account.first.cname
2.7.4 :003 > ActiveFedora::SolrService.instance.conn.options
 => {"read_timeout"=>120, "open_timeout"=>120, "url"=>"http://...", "collection"=>"UUID"}
```
I think the issue is likely related to GCloud not starting the pods correctly.

PR #589, opened 2021-12-21
Hotfix/doi job rescue

BertZZ:


app/jobs/concerns/hyku_addons/register_doi_job_override.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  module RegisterDOIJobOverride
+    extend ActiveSupport::Concern
+
+    included do
+      rescue_from Hyrax::DOI::DataCiteClient::Error do |exception|
+        Rails.logger.debug exception.inspect
+        retry_job(wait: 5.minutes)
```
thenapking, line 10: Is it worth only retrying the job a limited number of times? I am concerned that this will flood the queue with jobs which will never succeed.  I am more familiar with Resque.  Apparently for Sidekiq you need to add custom middleware to do this.  Please see [this](https://stackoverflow.com/questions/23065578/sidekiq-retry-count-in-job) stack-exchange post which will give all the details.

BertZZ:
Not needed

PR #590, opened 2021-12-22
Add tests for JsonFieldsActor and CollectionForm

thenapking:
Add two tests to allow for refactoring of JsonFieldsActor into a service class.
Please comment on
1) blank/empty hashes appear not to be removed from array parameters, when one item in the array is not blank
2) Should there be examples around keys like "creator_name_type"? These appear to be filtered in some circumstances I am unclear about
3) Are the terms included in the CollectionForm spec correct
4) What behaviour should I have tested that is missing - I am not familiar with the purpose of these classes.

### DIFF
--- a/spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
+++ b/spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
@@ -33,6 +33,10 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "passes non-json fields through" do
         expect(env.attributes[:title]).to eq(transformed_attributes["title"])
       end
+
+      it "sets the attributes of the work" do
+        expect(work.title).to eq(transformed_attributes["title"])
+      end
     end
 
     context "with multiple valid json fields" do
@@ -56,6 +60,11 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "adds them all" do
         expect(env.attributes).to eq(transformed_attributes)
       end
+
+      it "sets the attributes of the work" do
+        expect(work.creator).to eq(transformed_attributes["creator"])
+        expect(work.editor).to eq(transformed_attributes["editor"])
+      end
     end
 
     context "with blank or empty attributes" do
@@ -77,6 +86,11 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "turns fields into null or adds empty arrays" do
         expect(env.attributes).to eq(transformed_attributes)
       end
+
+      it "sets the attributes of the work" do
+        expect(work.creator).to be_empty
+        expect(work.editor).to eq(transformed_attributes["editor"])
+      end
     end
 
     context "with no name type" do
@@ -94,6 +108,10 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
 
       it "adds the attributes" do
         expect(env.attributes).to eq(transformed_attributes)
+      end
+
+      it "sets the attributes of the work" do
+        expect(work.editor).to eq(transformed_attributes["editor"])
       end
     end
 
@@ -121,6 +139,10 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "recurses through them" do
         expect(env.attributes).to eq(transformed_attributes)
       end
+
+      it "sets the attributes of the work" do
+        expect(work.creator).to eq(transformed_attributes["creator"])
+      end
     end
   end
 
@@ -139,6 +161,10 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
 
       it "passes non-json fields through" do
         expect(env.attributes[:title]).to eq(transformed_attributes["title"])
+      end
+
+      it "updates the attributes of the work" do
+        expect(work.title).to eq(transformed_attributes["title"])
       end
     end
 
@@ -175,6 +201,11 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "adds them all" do
         expect(env.attributes).to eq(transformed_attributes)
       end
+
+      it "updates the attributes of the work" do
+        expect(work.creator).to eq(transformed_attributes["creator"])
+        expect(work.editor).to eq(transformed_attributes["editor"])
+      end
     end
 
     context "with blank or empty attributes" do
@@ -214,6 +245,11 @@ RSpec.describe HykuAddons::Actors::JSONFieldsActor do
       it "turns fields into null or adds empty arrays" do
         expect(env.attributes).to eq(transformed_attributes)
       end
+
+      it "updates the attributes of the work" do
+        expect(work.creator).to be_empty
+        expect(work.editor).to eq(transformed_attributes["editor"])
+      end
     end
 
     context "with no name type" do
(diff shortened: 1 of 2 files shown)


spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+      let(:transformed_attributes) do
+        { "editor" => ["[{\"editor_name_type\":\"\",\"editor_given_name\":\"Lihong V.\",\"editor_family_name\":\"Wang\"}]"], "creator" => [], "contributor" => [], "funder" => [], "alternate_identifier" => [], "related_identifier" => [], "current_he_institution" => [] }
+      end
+
+      it "adds the attributes" do
+        expect(env.attributes).to eq(transformed_attributes)
+      end
+    end
+
+    context "with nested fields" do
```
mankind, line 100: I am not sure we have this use case of nesting creator but it is okay to have the test.

spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+
+      it "turns fields into null or adds empty arrays" do
+        expect(env.attributes).to eq(transformed_attributes)
+      end
+    end
+
+    context "with no name type" do
+      let(:original_attributes) do
+        {
+          editor: [
```
mankind, line 222: Here the hash key is **editor** should the sub keys not be prefixed with editor instead of creator rg
  
      editor_name_type

as against

    creator_name_type

spec/forms/hyrax/collection_form_spec.rb
```diff
@@ -0,0 +1,85 @@
...
+    context ".model_attributes" do
+      let(:params) do
+        ActionController::Parameters.new(
+          title: [""],
+          license: [""],
+          related_url: [""],
+          keyword: ["not_blank"],
+          creator: [
+            { "creator_name_type" => "Personal", "creator_given_name" => "Alice", "creator_family_name" => "Coltrane" },
+            {} # IS THIS CORRECT ?? IT IS NOT REMOVED
```
mankind, line 69: I can't remember if we have a situation where empty hash is in the list of hashes in **creator**  because I think the javascript in the views eliminates it, but if we think that we can have empty hash in the data when coming from elsewhere eg imports, then may be the json_field_actor refactor should ensure it is not saved.

spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+          creator: [
+            [{
+              creator_name_type: "Personal",
+              creator_given_name: "Johnny",
+              creator_family_name: "Testison"
+            }].to_json
+          ],
+          editor: [
+            [{
+              creator_name_type: "Personal",
```
mankind, line 192: subkeys should have **editor **prefix not **creator**

spec/forms/hyrax/collection_form_spec.rb
thenapking: Cool.  It seemed pointless to me to have this code to remove empty hashes etc which doesn't actually do it.

spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
thenapking: I wanted to test the recursion. Can you think of an example where this might actually get used?

mankind: If you mean an example where we might use nested creator, it is hard to think of one since the journals from where we get the creators don't have nested creators in as part of paper publications.

Also in over 2 years since this json feature was implemented and in use in the wild, that use case has not come up.

mankind:
I have a few general comments:

- I keep seeing **original_attributes ** inside each context but beside their definition, it is not clear what their use is there. The **env** always takes in the **attributes** and checks **transformed_hash**.

thenapking:
> I have a few general comments:
> 
> * I keep seeing **original_attributes ** inside each context but beside their definition, it is not clear what their use is there. The **env** always takes in the **attributes** and checks **transformed_hash**.

I may have this setup wrong. I wanted to create an object with some original attributes, and then test that those attributes are updated with the new attributes.  Is this the wrong way to test that?

spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+        { "creator" => ["[{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Liang\",\"creator_family_name\":\"Gao\"},{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Lidai\",\"creator_family_name\":\"Wang\"},{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Chiye\",\"creator_family_name\":\"Li\"},{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Yan\",\"creator_family_name\":\"Liu\"},{\"creator\":[{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Haixin\",\"creator_family_name\":\"Ke\"},{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Chi\",\"creator_family_name\":\"Zhang\"},{\"creator_name_type\":\"Personal\",\"creator_given_name\":\"Lihong V.\",\"creator_family_name\":\"Wang\"}]}]"], "contributor" => [], "funder" => [], "alternate_identifier" => [], "related_identifier" => [], "editor" => [], "current_he_institution" => [] }
+      end
+
+      it "recurses through them" do
+        expect(env.attributes).to eq(transformed_attributes)
+      end
+    end
+  end
+
+  describe "#update" do
```
mankind, line 127: I am not sure this is testsing that the work is updated based on the values passed to **env**.  Actor update is supposed to update the actual **work** passed to it with the **attributes** passed to it, so should the expectation not test that the work creator was updated to match the attributes passed to env.

mankind:
> > I have a few general comments:
> > 
> > * I keep seeing **original_attributes ** inside each context but beside their definition, it is not clear what their use is there. The **env** always takes in the **attributes** and checks **transformed_hash**.
> 
> I may have this setup wrong. I wanted to create an object with some original attributes, and then test that those attributes are updated with the new attributes. Is this the wrong way to test that?

If we look at this lines  below for example, it should check that he values of the **work creator** were now same as the attributes passed to the env which is different from what was saved to the work using **original_attributes**. The current expectation checks that is was transformed currently into json and i think we should keep that  test too.

https://github.com/ubiquitypress/hyku_addons/blob/93d8ed3ac1586e9e144f57124abe993810040b12/spec/actors/hyku_addons/actors/json_fields_actor_spec.rb#L145-L178

spec/actors/hyku_addons/actors/json_fields_actor_spec.rb
thenapking: OK.  So it sounds like I should create the work itself, outside of the env, then call update on it and test the attributes have been updated correctly.

mankind: I think what you already have the right setup, just to check in the expectation that the work json fields changed in line new the json fields values passed to the actor

spec/forms/hyrax/collection_form_spec.rb
thenapking: I will update this test as part of the refactor.  I will update the criteria so that empty hashes are removed.

mankind (approved):
Thank you for the changes :+1:

PR #591, opened 2021-12-23
Further Author profile work 

BertZZ:
Additional work for the Author profile. 

- Adds Biography field to the user model and Biography and Department fields to user form
- Hides public profile Toggle unless Admin
- Adds creator_institutional_email to creator block to link users to their works 
- Returns works where creator_institutional_email matches user email as part of user endpoint response
- Removes creator_institutional_email from creator block in the work response if it is blank or the profile is private/ not found so that the fronted knows when to link to author profile from Show Work Page.


app/controllers/concerns/hyku_addons/users_controller_behavior.rb
```diff
@@ -12,7 +12,15 @@ def index
 
     def show
       @user = User.find_by(email: params[:email], display_profile: true)
-      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
+      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } and return if @user.blank?
```
prdanelli, line 15: This is pretty complicated: 

```
if @user.blank?
    return render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } 
end
````

Comment (BertZZ) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: Your suggestion fails RuboCop::Cop::Style::IfUnlessModifier

Comment (prdanelli) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: I dont know what that means. What is the suggested alternative?

Comment (BertZZ) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/IfUnlessModifier

Basically exactly as I have it

Comment (prdanelli) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: My issues wasn't with the end of line `if/unless`, it was the `and return`. 

Why not `return render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?` ?

Comment (BertZZ) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: Ill give it a go

Comment (prdanelli) [reply app/controllers/concerns/hyku_addons/users_controller_behavior.rb:15]: 'do_soemthing and return` is different from `return do_something`.

File: app/views/hyku/api/v1/work/_work.json.jbuilder [hunk 2]
```diff
@@ -23,10 +23,12 @@ json.cache! [@account, :works, work.id, work.solr_document[:_version_], work.mem
 
   creator_hash = work.creator.try(:first)
   creator_json = JSON.parse(creator_hash) if creator_hash.present?
-  creator_json.each_with_index do |creator, index|
-    creator_json[index] = creator_json[index].except!("creator_institutional_email") && next unless creator["creator_institutional_email"].present? # Removes field if blank
-    user = User.find_by(email: creator["creator_institutional_email"])
-    creator_json[index] = creator_json[index].except!("creator_institutional_email") unless user.present? && user.display_profile # Removes field if the user is not public/not found
+  if creator_json.present?
+    creator_json.each_with_index do |creator, index|
+      creator_json[index] = creator_json[index].except!("creator_institutional_email") && next unless creator["creator_institutional_email"].present? # Removes field if blank
```
prdanelli, line 28: You're doing too much on one line here, assigning a variable, moving to the next iteration and checking present. Its ok to use more than one line if you are going multiple things and it makes the code easier to read. A giant block of code needs to be spaced to that the next developer that comes to it can process it quickly. 

Also, avoid `unless present?` if you can use `if blank?`. 

```
if creator["creator_institutional_email"].blank?
    creator_json[index] = creator_json[index].except!("creator_institutional_email") 
    next 
end
```

.rubocop.yml
```diff
@@ -139,6 +139,10 @@ Style/SymbolProc:
   Exclude:
     - 'app/models/concerns/hyku_addons/account_behavior.rb'
 
+Style/AndOr:
```
prdanelli, line 142: You dont need this anymore. 

It's also a bad idea to exclude an entire file from being checked. You should use the comments `# rubocop:disable Rulename` and ` # rubocop:enable Rulename` around the block you are working with.

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
```diff
@@ -12,7 +12,15 @@ def index
 
     def show
       @user = User.find_by(email: params[:email], display_profile: true)
-      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
+      return render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
+
+      query_string = "(creator_tesim:\"*#{@user.email}*\") AND visibility_ssi:open"
+      @user_document_id_list = ActiveFedora::SolrService.get(query_string, rowy: 1_000_000)['response']['docs'].pluck('id')
```
prdanelli, line 18: Is the argument supposed to be called `rowy`?

BertZZ: I think it actually is but I will double check

BertZZ: I thinks it's probably rows isn't it. Ill change it

prdanelli: Can you give your PR's a read through before you give them to anyone else, just incase you have little things like this in them that you can catch and fix up :)

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
thenapking: To clarify the situation: if you don't add a return statement here do you get a double render error? If that's the case instead of using `render {} and return` you should try to construct this so you don't double render.  I presume this is happening because you specifically want to render an error message and return a 403.  In that case you need to correctly override the other bit of code that calls render so that it is not called if `@user.blank?`.

app/views/hyku/api/v1/work/_work.json.jbuilder
```diff
@@ -21,8 +21,20 @@ json.cache! [@account, :works, work.id, work.solr_document[:_version_], work.mem
...
+      if creator["creator_institutional_email"].blank?
+        creator_json[index] = creator_json[index].except!("creator_institutional_email")
+        next
+      else
+        user = User.find_by(email: creator["creator_institutional_email"])
+        creator_json[index] = creator_json[index].except!("creator_institutional_email") unless user.present? && user.display_profile # Removes field if the user is not public/not found
+      end
+    end
+  end
+  json.creator creator_hash.present? ? creator_json : []
```
prdanelli, line 37: This block is a little muddled. You're creating the `creator_hash` at the top, then checking the hash to create the json, processing the json then checking the hash again at the bottom. 

You shouldn't be doing this kind of querying inside a view really, the view should be as dumb as possible - I know that with Hyku its harder than it should be sometimes, but this kind of thing could be encapsulated inside a Helper and then just written as: 

```ruby
json.creator processed_creator(work.creator.try(:first))
```

Big sections of code like this can become hard to comprehend, so its best to not condense them down and remove the line breaks. If you space it out, it'll be easier for you to see where you can make improvements. 

Your variables are the wrong names. `creator_hash = work.creator.try(:first)` will return JSON, and `creator_json = JSON.parse(creator_hash)` will return a hash. 

The method `except!` changes the actual hash that you are on, so you dont need to assign that to another hash. So instead of: 
```ruby
creator_json[index] = creator_json[index].except!("creator_institutional_email")
```
You could just write 
```ruby
creator_json[index].except!("creator_institutional_email")
```
For example: 
```
2.7.4 :004 > h = {one: 1, two: 2}
 => {:one=>1, :two=>2}
2.7.4 :005 > h.except!(:one)
 => {:two=>2}
2.7.4 :006 > h
 => {:two=>2}
```
But you dont even need to do that as the `delete!` method will remove the requested key from the array. So, you can just use `creator_json[index].delete("creator_institutional_email")`. 

For example: 
```
2.7.4 :001 > h = {one: 1, two: 2}
 => {:one=>1, :two=>2}
2.7.4 :002 > h.delete(:one)
 => 1
2.7.4 :003 > h
 => {:two=>2}
```

You're using `creator_json.each_with_index do |creator, index|` and then accessing the `creator[index]`, but why not just use `map`?

I've not tested this, but adding this as  helper method would be preferable to the block you added to the view: 

```ruby
def processed_creator(creator_json)
  creator_hash = JSON.parse(creator_json || "{}") 

  return if creator_hash.blank?

  key = "creator_institutional_email"

  creator_hash.map do |creator|
    if creator[key].blank?
      creator.delete(key)
    else
      user = User.find_by(email: creator[key])
      creator_json[index].delete(key) unless user.present? && user.display_profile 
    end
  end
end
```

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
BertZZ: I don't think there was actually a double render error as I thought originally 🙈🙈. I think i remembered using the `and return` in the past to stop code running and just applied it here. 

The actual reason to return is so that we don't call `@user.email ` if there is no user

prdanelli (changes requested):

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
```diff
@@ -12,7 +12,15 @@ def index
...
-      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
+      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } and return if @user.blank?
+
+      query_string = "(creator_tesim:\"*#{@user.email}*\") AND visibility_ssi:open"
+      @user_document_id_list = ActiveFedora::SolrService.get(query_string, rowy: 1_000_000)['response']['docs'].pluck('id')
+      @user_works = Hyrax::PresenterFactory.build_for(ids: @user_document_id_list,
+                                                      presenter_class: Hyku::WorkShowPresenter,
+                                                      presenter_args: current_ability)
+      collection_search_builder = Hyrax::CollectionSearchBuilder.new(self).with_access(:read).rows(1_000_000)
+      @collection_docs = repository.search(collection_search_builder).documents
```
prdanelli, line 23: Are you using all three `@user_document_id_list`, `@user_works` and `@collection_docs` in the view or in other methods inside of the request?

The entire block of code could be moved to a seperate protected method to reduce the noise and if only a single instance variable is being used to pass data to the view, then the others can be made normal method variables, which means the garbage collection could potentially pick them up faster and reduce the app memory usage.

BertZZ: I am not super sure :sweat_smile: I found a similar block of code in the engine file and used that. I will add it to the list of things to look at and refactor :smiley:

prdanelli: You should try not to just copy and paste. Writing the code out by hand might seem more long winded, but you get several benefits from doing it. You'll remember the code many times quicker than copying/pasting it and you'll notice what is and is not required for your specific context.

BertZZ: `@user_works` and `@collection_docs` are both used in the view. I don't think its a good idea to assign instance variables in protected methods is it?

app/views/hyrax/dashboard/profiles/_edit_primary.html.erb
```diff
@@ -19,12 +19,14 @@
     <br>
     <!-- I would like to move this to the bottom of the form but whenever I try the styling is all
     wrong and I don't want to push the pixels around right now-->
+    <% if current_user.has_role?(:admin, Site.instance) %>
      <div class="checkbox">
        <%= f.label :allow_public_display do %>
         <%= f.check_box :display_profile %>
```
prdanelli, line 25: The label symbol and the checkbox symbol need to match otherwise you'll break some accesibility rules in HTML. 

```
       <%= f.label :public_display do %>
        <%= f.check_box :display_profile %>
```

BertZZ: Thanks :)

app/helpers/hyku_addons/api_works_helper.rb
```diff
@@ -8,16 +8,13 @@ def creator_from_json(creator_json)
 
       return if creator_hash.blank?
 
-      key = "creator_institutional_email"
```
prdanelli: You've remove the variable and are now repeating your variable string multiple times.

app/helpers/hyku_addons/api_works_helper.rb
```diff
@@ -8,16 +8,13 @@ def creator_from_json(creator_json)
...
-
       creator_hash.map do |creator|
-        if creator[key] && creator[key].blank?
-          creator.delete(key)
+        if creator["creator_institutional_email"]&.blank?
```
prdanelli, line 12: This isn't the same as the code that was there before. The previous code checked that the index was set and then that the value was "blank". You dont need to use `&` because `blank?` responds to `nil` and any other type that you will be dealing with here.

prdanelli (changes requested):
By remove the `key` variable you've made the code noisier than it needs to be. By using a variable you keep your code DRY (dont repeat yourself).

app/helpers/hyku_addons/api_works_helper.rb
BertZZ: And Rubocop is no longer shouting at me about complexity. You rather I just ignore the Cop?

app/helpers/hyku_addons/api_works_helper.rb
BertZZ: ok I think maybe we need to figure out a better way to write it to please rubocop and do what we want. Ill take a look on Monday

app/helpers/hyku_addons/api_works_helper.rb
prdanelli: Rubocop isn't God :) 

But you shouldn't make code worse just to appease it. Rubocop uses the number of `.`'s (and other stuff like if statements) to determine the potential complexity. So by moving the `user` query out to a helper method you can also save complexity there.

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
```diff
@@ -12,7 +12,15 @@ def index
 
     def show
       @user = User.find_by(email: params[:email], display_profile: true)
-      render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
+      return render json: { status: 403, code: 'forbidden', message: t("errors.users_forbidden") } if @user.blank?
```
prdanelli, line 15: Can you also make sure you're using double quotes, they are both being used through the method.

app/helpers/hyku_addons/api_works_helper.rb
```diff
@@ -0,0 +1,28 @@
...
+      return [] if creator_hash.blank?
+
+      key = "creator_institutional_email"
+
+      creator_hash.map do |creator|
+        if creator[key].blank?
+          creator.delete(key)
+        else
+          user = find_by_user_email(creator[key])
+          creator.delete(key) unless user.present? && user.display_profile
```
prdanelli, line 18: This looks much clearer now, but there i a small improvement we can make. Instead of adding in another method to find the user, then check for the object and a value on it, we can move all of that logic to the method. 

```ruby
    # ...
    creator.delete(key) unless user_display_profile?(creator[key])
    # ....

    def user_display_profile?(email)
        user = User.find_by(email: email)

        user.present? && user.display_profile
    end
````
But now that we have it condensed to a single line, we can improve it once more... 
```ruby
    creator_hash.map do |creator|
        if creator[key].blank? 
            creator.delete(key)
        elsif !user_display_profile?(creator[key])
            creator.delete(key)
        end
    end
```
We can now see that we have the same statment being run from two conditions, so there is another improvement we can make: 
```ruby
    creator_hash.map do |creator|
        creator.delete(key) if creator[key].blank? || !user_display_profile?(creator[key])
    end
```
But, its not great using an inline `if` like this for a single line, when we can invert the condition and use a guard statement instead: 
```ruby
    creator_hash.map do |creator|
         next if creator[key].present? && user_display_profile?(creator[key])
 
         creator.delete(key)
    end
```
There is one downside to this and that is because the guard clause uses an `&&`, both sides of the condition will need to be evaluated, which means we are going to do a db query for each iteration. This could potentially cause issues with larger sets of user emails. 

By using an `||` if the left side of the condition is true, then we do not need to move onto the left side: 
```ruby
    creator_hash.map do |creator|
         # Use unless to avoid querying the database for each user
         next unless creator[key].present? || !user_display_profile?(creator[key])
 
         creator.delete(key)
    end
```

BertZZ: Should the last one be 
`next unless creator[key].blank? || !user_display_profile?(creator[key])`

prdanelli: Quite possibly. Why do you think that? You know the end goal of this code block better than I do. 

So the original was:
```ruby
        if creator[key].blank?
          creator.delete(key)
        else
          user = find_by_user_email(creator[key])
          creator.delete(key) unless user.present? && user.display_profile
        end
```
and we ended up with: 
```ruby
    creator_hash.map do |creator|
         next unless creator[key].present? || !user_display_profile?(creator[key])
 
         creator.delete(key)
    end
```
So if we want to ignore the creator if their `creator_institutional_email` is available, or it doesn't relate to a user in the system, whilst also avoiding querying the db every iteration. 

```ruby
    creator_hash.map do |creator|
         next if creator[key].present? || !user_display_profile?(creator[key])
 
         creator.delete(key)
    end
```

prdanelli (commented):
I have left some comments, but its looking good. Just need to iron out some bits.

app/helpers/hyku_addons/creator_field_helper.rb
```diff
@@ -7,15 +8,15 @@ def add_pacific_creator_personal_fields(array)
     end
 
     def add_redlands_creator_personal_fields(array, service_options)
-      array.delete_at(6) # removes institutional relationship from Redlands worktype
+      array.delete_at(7) # removes institutional relationship from Redlands worktype
```
prdanelli, line 11: Is this from your PR, or is this an outdated change?

BertZZ: This is due to the addition of the `creator_instiutional_email` field to the array

app/actors/hyku_addons/actors/creator_institutional_email_actor.rb
```diff
@@ -18,12 +18,13 @@ def check_creator_institutional_email(env)
           return [] if creator_hash.blank?
 
           creators = creator_hash.map do |creator|
+            next if creator["creator_name_type"] == "Organizational"
```
prdanelli, line 21: Don't forget to add a break return after guard statements to ensure they are easily seen by other developers reading the code

app/actors/hyku_addons/actors/creator_institutional_email_actor.rb
```diff
@@ -0,0 +1,31 @@
...
+
+      private
+
+        def check_creator_institutional_email(env)
+          creator_hash = JSON.parse(env.attributes[:creator].try(:first) || "{}")
+
+          return [] if creator_hash.blank?
+
+          creators = creator_hash.map do |creator|
+            return creator if creator["creator_name_type"] == "Organizational"
```
prdanelli, line 21: Don't forget to add a break return after your guard statements to clearly illustrate what is happening.

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
```diff
@@ -12,7 +12,15 @@ def index
 
     def show
       @user = User.find_by(email: params[:email], display_profile: true)
```
prdanelli, line 14: Is `@user` used in the view? If not, it could be a normal method variabe.

app/jobs/hyku_addons/toggle_display_profile_job.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  class ToggleDisplayProfileJob < ApplicationJob
+    def perform(email, display_profile)
+      query_string = "(creator_tesim:\"*#{email}*\")"
+      works = ActiveFedora::SolrService.get(query_string, rows: 1_000_000)["response"]["docs"]
```
prdanelli, line 7: Ideally, if you are chaining methods like this, you would use `dig`, so: 

```ruby
works = ActiveFedora::SolrService.get(query_string, rows: 1_000_000)
                                 .dig("response", "docs")
                                 .map { |doc| ActiveFedora::SolrHit.new(doc).reify }
```

app/controllers/concerns/hyku_addons/users_controller_behavior.rb
BertZZ: https://github.com/ubiquitypress/hyku-api/blob/main/app/views/hyku/api/v1/users/show.json.jbuilder

app/actors/hyku_addons/actors/creator_institutional_email_actor.rb
```diff
@@ -0,0 +1,32 @@
...
+
+          creators = creator_hash.map do |creator|
+            if %w[Organizational Organisational].include?(creator["creator_name_type"])
+            else
+              user = User.find_by(email: creator["creator_institutional_email"])
+              creator["display_creator_profile"] = user.present? && user.display_profile
+            end
+            creator
+          end
+          env.attributes[:creator] = [creators.to_json]
```
prdanelli, line 28: Can you add a break return above this line to make it clear that there is no relation between these lines.

prdanelli (dismissed):
Looks good.

PR #595, opened 2022-01-06
Resolve issues with the note field for schema works

prdanelli:
Because the note field had its symbol added to the permitted params via the normal method, but was included in a seperate tab, it cannot be included in the schema YAML as the other fields are. This is the case for all special case fields, Hyrax-DOI fields, Notes, etc that need their own form tabs. If they are included in the schema definition, those values with override the field within the tab and vice versa.


mankind (approved):
:+1:

PR #603, opened 2022-01-11
Update Gemfile.lock from inside the linux container

prdanelli:
Whenever the `bundle install` is run from outside of the container, the Google Gems are compiled for the Host OS, then when the `bundle install` in run from inside the containers it compiles for native linux. We need to only ever bundle install from inside of the container and if you do bundle install from your host machine, those changes should not be committed.


thenapking (approved):
LGTM

PR #605, opened 2022-01-12
Test that DOIs are minted by Bulkrax import

thenapking:
Update CSV fixtures
Add tests which allow Bulkrax import to trigger `Hyrax::DOI::RegisterDOIJob`
Add tests which demonstrate that Bulkrax can set statuses for new works
Add tests which demonstrate that Bulkrax can set statuses for existing works
Add code to ensure that Bulkrax does not move DOI status backwards for existing jobs by comparing parsed metadata with existing records

NB: This means that the CsvParser class now is coupled to the record, and will look up the record if the doi_status_when_public attribute is present in the metadata.  Another approach might be to add a validation which prevents this status moving backwards.  However this would require some complex logic regarding when to apply this validation.


app/models/hyku_addons/csv_entry.rb
```diff
@@ -210,5 +221,14 @@ def hyrax_record
       rescue
       end || factory.find
     end
+
+    private
+
+      def current_doi_status
+        query = { Bulkrax.system_identifier_field => record['source_identifier'] }
+        match = factory_class.where(query).detect { |m| m.send(Bulkrax.system_identifier_field).include?(record['source_identifier']) }
```
prdanelli, line 229: Can we change the single to double quotes please?

spec/features/bulkrax_import_spec.rb
```diff
@@ -15,13 +15,6 @@
            limit: 0)
   end
   let(:import_batch_file) { "spec/fixtures/csv/pacific_articles.metadata.csv" }
-  # let(:field_mapping) do
```
prdanelli, line 18: Love this. 10/10

prdanelli (commented):
Just a little change requested :)

spec/features/bulkrax_import_spec.rb
```diff
@@ -281,6 +274,153 @@
...
+            expect(work_4.reload.doi).to eq([doi])
+
+            expect(work_1.doi_status_when_public).to be_nil
+            expect(work_2.doi_status_when_public).to eq("draft")
+            expect(work_3.doi_status_when_public).to eq("registered")
+            expect(work_4.doi_status_when_public).to eq("findable")
+          end
+        end
+
+        context "the import changes the DOI status backwards" do
```
mankind, line 398: By taking doi backwards do yo mean to what is  in the imported csv as against the values in work-1 to work-4?

spec/features/bulkrax_import_spec.rb
```diff
@@ -281,6 +274,153 @@
...
+          let!(:work_1) { create :work, source_identifier: "external-id-1", doi: [doi], doi_status_when_public: "draft" }
+          let!(:work_2) { create :work, source_identifier: "external-id-2", doi: [doi], doi_status_when_public: "registered" }
+          let!(:work_3) { create :work, source_identifier: "external-id-3", doi: [doi], doi_status_when_public: "findable" }
+          let!(:work_4) { create :work, source_identifier: "external-id-4", doi: [doi], doi_status_when_public: "findable" }
+
```
mankind, line 403: By taking doi backwards do yo mean to what is  in the imported csv as against the values in work-1 to work-4?

thenapking: Yes, so there are two tests here which could be rolled into one.  
(a) the import does not change the DOI status
(b) the import does not change the DOI status backwards (my mistake corrected)
Since there are 4 rows in the CSV (I could add more....) I think one test could cover both of these current tests.

mankind: @thenapking is failing on codeclimate, should i approve, since it is not related to your code and i have not seen this failure before, hence might be something flaky.

mankind: Sounds good :+1:

thenapking: I have improved the specs here, if you would like to re-review: [a7ce3d1](https://github.com/ubiquitypress/hyku_addons/pull/605/commits/a7ce3d10626e1ac1e6d14ccba2f1848455f3d662)

app/models/hyku_addons/csv_entry.rb
```diff
@@ -1,113 +1,17 @@
...
-      elsif record['file'].is_a?(Array)
-        parsed_metadata['file'] = record['file']
-      end
-      parsed_metadata['file'] = parsed_metadata['file'].map { |f| path_to_file(f) }
-    end
-
     # Override to allow `id` as system identifier field
     def valid_system_id(model_class)
-      return true if Bulkrax.system_identifier_field == 'id'
+      return true if Bulkrax.system_identifier_field == "id"
```
prdanelli, line 14: 🥇

prdanelli (approved):
Can't pretend I understand much of it, but you seem to have it pretty well covered. Good work.

PR #607, opened 2022-01-14
Fix Flaky Account Settings tests

thenapking:
Ensure that the `:host` parameter is removed from `default_url_options` to prevent Account objects created in one test leaking into others.


prdanelli (approved):
Thanks for sticking with it and unbreaking the tests ;)

PR #608, opened 2022-01-14
Dashboard GDS reports

prdanelli:
MVP for displaying Google Data Studio reports on a new page in the dashboard. 

image

Config is saved within the account setting section as a formatted list of chart URLs to iterate over in the view. 

image

The option can be enabled with the Feature Flipper. 

image

### DIFF
--- a/app/assets/stylesheets/hyku_addons/dashboard.scss
+++ b/app/assets/stylesheets/hyku_addons/dashboard.scss
@@ -76,6 +76,7 @@
     max-width: 100%;
   }
 }
+
 .btn {
   &[data-create-type="batch"], &[data-behavior="batch-edit"] {
     display: none;
@@ -112,6 +113,18 @@ ul.listing {
           text-decoration: underline;
         }
       }
+    }
+  }
+}
+
+#reports {
+  #dashaboard_gds_reports {
+    margin: 10px 0 20px;
+  }
+
+  #report {
+    iframe {
+      display: none;
     }
   }
 }
--- a/app/assets/stylesheets/hyku_addons/dashboard.scss
+++ b/app/assets/stylesheets/hyku_addons/dashboard.scss
@@ -1,3 +1,81 @@
+// Add the boostrap 4 .col-xl- breakpoint to the dashboard as version 3 only has .col-lg
+@media only screen and (min-width: 1600px) {
+  .col-xl-1, .col-xl-2, .col-xl-3, .col-xl-4, .col-xl-5, .col-xl-6, .col-xl-7, .col-xl-8, .col-xl-9, .col-xl-10, .col-xl-11, .col-xl-12, .col-xl,
+  .col-xl-auto {
+    position: relative;
+    width: 100%;
+    padding-right: 15px;
+    padding-left: 15px;
+    float: left;
+  }
+
+  .col-xl-auto {
+    -ms-flex: 0 0 auto;
+    flex: 0 0 auto;
+    width: auto;
+    max-width: 100%;
+  }
+  .col-xl-1 {
+    -ms-flex: 0 0 8.333333%;
+    flex: 0 0 8.333333%;
+    max-width: 8.333333%;
+  }
+  .col-xl-2 {
+    -ms-flex: 0 0 16.666667%;
+    flex: 0 0 16.666667%;
+    max-width: 16.666667%;
+  }
+  .col-xl-3 {
+    -ms-flex: 0 0 25%;
+    flex: 0 0 25%;
+    max-width: 25%;
+  }
+  .col-xl-4 {
+    -ms-flex: 0 0 33.333333%;
+    flex: 0 0 33.333333%;
+    max-width: 33.333333%;
+  }
+  .col-xl-5 {
+    -ms-flex: 0 0 41.666667%;
+    flex: 0 0 41.666667%;
+    max-width: 41.666667%;
+  }
+  .col-xl-6 {
+    -ms-flex: 0 0 50%;
+    flex: 0 0 50%;
+    max-width: 50%;
+  }
+  .col-xl-7 {
+    -ms-flex: 0 0 58.333333%;
+    flex: 0 0 58.333333%;
+    max-width: 58.333333%;
+  }
+  .col-xl-8 {
+    -ms-flex: 0 0 66.666667%;
+    flex: 0 0 66.666667%;
+    max-width: 66.666667%;
+  }
+  .col-xl-9 {
+    -ms-flex: 0 0 75%;
+    flex: 0 0 75%;
+    max-width: 75%;
+  }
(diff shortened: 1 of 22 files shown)


app/assets/stylesheets/hyku_addons/dashboard.scss
```diff
@@ -1,3 +1,82 @@
+// Add the boostrap 4 .col-xl- breakpoint to the dashboard as version 3 only has .col-lg
```
mankind, line 1: If we are on version 3 of Bootstrap, is there a reason why we need Bootstrap 4 css classes for this?

app/views/hyku_addons/account_settings/index.html.erb
```diff
@@ -10,41 +10,59 @@
   <tbody>
     <% HykuAddons::AccountBehavior::TEXT_SETTINGS.each do |fname| %>
       <tr>
-        <td id="link-<%= fname %>"><b><%= fname.humanize %>:</b> <%= @account&.settings[fname] %></td>
-        <td><%= link_to "Edit #{fname.humanize}", hyku_addons.edit_admin_account_setting_path(id: @account, partial_name: 'render_text_fields', field_name: fname), class: "btn btn-xs btn-default", remote: true %></td>
+        <% title_key = "settings.titles.#{fname}" %>
```
mankind, line 13: This has been repeated almost 4 times:

            <% title_key = "settings.titles.#{fname}" %>

I know it is called from within different each block but do you think that using a helper will clean it up.

app/assets/stylesheets/hyku_addons/dashboard.scss
prdanelli: Its only the `xl` grid styles, which are actually taken from Bootstrap 4, because bootstrap 3 doesn't support any breakpoints above 1200, which is ridiculous in 2022. So we now have the ability to customise the layout just a little bit more for screens over 1600 and everything else falls back to the defaults.

spec/controllers/hyrax/stats_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
+# frozen_string_literal: true
+
+require "rails_helper"
+
+RSpec.describe Hyrax::StatsController, type: :controller do
+  routes { HykuAddons::Engine.routes }
+
+  let(:user) { create(:admin) }
+  let(:account) { create(:account) }
```
thenapking, line 9: can we use `build_stubbed` or does this need `create`?

app/views/hyku_addons/account_settings/index.html.erb
prdanelli: You are absolute right Ed. I've moved the logic to the Helper.

spec/controllers/hyrax/stats_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+          get :reports
+        end
+
+        it "displays the content" do
+          expect(response.body).to include(I18n.t("stats.reports.title"))
+          expect(response.body).to include(I18n.t("stats.reports.intro"))
+        end
+
+        it "displays the select" do
+          expect(response.body).to have_selector("#dashaboard_gds_reports")
```
mankind, line 49: I am not sure testing selectors in controller test is the right place

spec/controllers/hyrax/stats_controller_spec.rb
prdanelli: For the `account` we can use `build_stubbed` but the users need to be records with their roles defined. I've updated the account however.

app/controllers/concerns/hyku_addons/stats_controller_behavior.rb
```diff
@@ -0,0 +1,35 @@
...
+        [config.delete_at(0), config.join(",")]
+      end
+    end
+
+    protected
+
+      def permission?
+        return if current_user.has_role?(:admin, Site.instance) && Flipflop.enabled?(:gds_reports)
+
+        raise ActionController::RoutingError, "Not found"
```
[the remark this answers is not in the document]
prdanelli: It's just seen as a 404 error by the user if they are given the URL and dont have permission. They shouldn't see the link at all, but its just to account for browser history or emailed links.

app/assets/stylesheets/hyku_addons/dashboard.scss
mankind: Ok thanks.

spec/controllers/hyrax/stats_controller_spec.rb
prdanelli: I dont love it either, which is why I was looking at system specs, but to add a whole new system spec for 2 expectations, which would take ages to get the permissions setup correctly for, seems like it wouldn't be worth it.

mankind: Thanks

mankind (dismissed):
:+1:

spec/controllers/hyrax/stats_controller_spec.rb
```diff
@@ -0,0 +1,88 @@
...
+          expect(response.body).to include(I18n.t("stats.reports.title"))
+          expect(response.body).to include(I18n.t("stats.reports.no_reports"))
+        end
+      end
+
+      context "when the user is a non-admin" do
+        let(:enabled) { false }
+
+        it "renders a partial" do
+          expect { get :reports }.to raise_error(ActionController::RoutingError)
```
thenapking, line 75: Does this routing error get rescued so the user is redirected?  Do we also need to test that a link a non-admin user can't access isn't displayed in a menu or such?

app/views/hyrax/stats/reports.html.erb
```diff
@@ -0,0 +1,48 @@
+<%= content_tag :h1, I18n.t("stats.reports.title"), class: "lower" %>
+
+<div id="reports">
+  <% if @reports.any? %>
+    <div class="row">
+      <div class="col-lg-3 col-md-4 col-sm-6 col-xs-12 mb-5 form-group">
+        <%= label_tag :dashaboard_gds_reports, "Available Reports", class: "form-label" %>
```
thenapking, line 7: Should this string also be in I18n?

spec/controllers/hyrax/stats_controller_spec.rb
prdanelli: I would suggest its fine for a user to see a 404 for a page that they are unable to access. As these are controller specs they're not really designed for testing out the content of a page, so for a non-admin user I would have to perform a GET to another controller action in order to see if the link they could click is hidden from them

app/views/hyrax/stats/reports.html.erb
prdanelli: Yes, yes it should.

PR #609, opened 2022-01-17
ensure tenant is switched before query is ran

mankind:
Collection banner url was not being returned in the api response because it was  wrongly assumed that the account tenant was already switched higher up in the request cycle so we did not need to do another switch at the point of running the query for collection_banner_info.

The test did not catch this because in the test we were switching tenant, so the fix was just to switch tenant before running the query.

### DIFF
--- a/app/services/hyku_addons/collection_branding_fetcher.rb
+++ b/app/services/hyku_addons/collection_branding_fetcher.rb
@@ -7,10 +7,10 @@ module HykuAddons
       @_collection_banner ||= collection_branding_info(id).where(role: 'banner')
     end
 
-    private_class_method
+    private
 
-    def collection_branding_info(id)
-      @_collection_branding_info ||= CollectionBrandingInfo.where(collection_id: id.to_s)
-    end
+      def collection_branding_info(id)
+        CollectionBrandingInfo.where(collection_id: id.to_s)
+      end
   end
 end
--- a/app/services/hyku_addons/collection_branding_fetcher.rb
+++ b/app/services/hyku_addons/collection_branding_fetcher.rb
@@ -2,14 +2,14 @@
 
 module HykuAddons
   class CollectionBrandingFetcher
-    def self.collection_banner(id, cname)
+    def collection_banner(id, cname)
       AccountElevator.switch!(cname)
       @_collection_banner ||= collection_branding_info(id).where(role: 'banner')
     end
 
     private_class_method
 
-    def self.collection_branding_info(id)
+    def collection_branding_info(id)
       @_collection_branding_info ||= CollectionBrandingInfo.where(collection_id: id.to_s)
     end
   end
--- a/app/services/hyku_addons/collection_branding_fetcher.rb
+++ b/app/services/hyku_addons/collection_branding_fetcher.rb
@@ -2,7 +2,8 @@
 
 module HykuAddons
   class CollectionBrandingFetcher
-    def self.collection_banner(id)
+    def self.collection_banner(id, cname)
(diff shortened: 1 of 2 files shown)


app/services/hyku_addons/collection_branding_fetcher.rb
```diff
@@ -2,7 +2,8 @@
 
 module HykuAddons
   class CollectionBrandingFetcher
-    def self.collection_banner(id)
+    def self.collection_banner(id, cname)
+      AccountElevator.switch!(cname)
       @_collection_banner ||= collection_branding_info(id).where(role: 'banner')
```
prdanelli, line 7: The class instance variable here i going to cause issues. Essentially, once set, it will not be unset until the server restarts as that is when the class is reloaded. 

```
2.7.4 :001 > class Foo
2.7.4 :002 >   def self.bar(number)
2.7.4 :003 >     @number ||= number
2.7.4 :004 >   end
2.7.4 :005 > end
 => :bar
2.7.4 :006 > foo = Foo.new
 => #<Foo:0x0000565351b6da28>
2.7.4 :007 > Foo.bar(1234)
 => 1234
2.7.4 :008 > Foo.bar(5678)
 => 1234
2.7.4 :009 > bar = Foo.new
 => #<Foo:0x0000565350282f40>
2.7.4 :010 > foo.class.instance_variable_get(:@number)
 => 1234
2.7.4 :011 > bar.class.instance_variable_get(:@number)
 => 1234
```
As you can see from this example, when set, all other instances of the the class will use whatever if the first value set, and all other times the class method is called will return the first value set. 

If you want to use a class method to set the banner, you cannot use a memoized variable inside of it.

prdanelli (changes requested):
I think you have an issue with the memoized class instance variable.

app/services/hyku_addons/collection_branding_fetcher.rb
mankind: Thanks @prdanelli, the code has been updated to use instance methods.

app/services/hyku_addons/collection_branding_fetcher.rb
```diff
@@ -2,13 +2,14 @@
...
     end
 
     private_class_method
 
-    def self.collection_branding_info(id)
+    def collection_branding_info(id)
       @_collection_branding_info ||= CollectionBrandingInfo.where(collection_id: id.to_s)
     end
```
prdanelli, line 14: So, i love memoizing complex functions, but is it worth it in this case? 

`CollectionBrandingFetcher` is only used in one json file and its assigned to a variable after which, the variable is used for the rest of the request. This just means that you will be storing the item in memory on the instance, then assigning it to another variable, so using double the memory that you might otherwise use if you weren't memoizing it. 

I think memoizing variables is great if you were calling the method multiple times, but as its only called once, its probably best to remove the extra compexity.

mankind: I think this is memoized because the query is complex and what is memoized in each is a different query response, so it is not thesame  data that is in memory

prdanelli: Sure, memoizing queries is a great idea, but this one only ever gets run one and then it assigned to a variable in the view, so the query is run, its stored on the instance, assigned to the view variable and that is accessed from then on. So there is never a secondary call of the method to make the memoization worthwhile.

mankind: I will remove one them.

PR #610, opened 2022-01-18
reindexmodeljob should skip retry & not mint when creator is blank

mankind:
During minting of doi when exceptions are raised, the work is re-queued.  The exception will still be handled but without retries.
Also if a work is sent for doi minting and it has no creator as was the case for some imported works, the minting will fail but in Datacite it will create a draft doi that is not attached to any work. We want to avoid that, by not minting doi for works that have no creator.

### DIFF
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -6,7 +6,7 @@
 
 module HykuAddons
   class ReindexModelJob < ApplicationJob
-    rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused do |exception|
+    rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused, Faraday::ConnectionFailed do |exception|
       Rails.logger.debug exception.inspect
     end
 
@@ -34,7 +34,7 @@ module HykuAddons
     private
 
       def mint_doi(work)
-        return if cant_mint_for?(work)
+        return unless cant_mint_for(work)
 
         Rails.logger.debug "=== about to mint doi for #{work.title} ==== "
 
@@ -43,8 +43,11 @@ module HykuAddons
         work.update(doi: [register_doi.identifier])
       end
 
-      def cant_mint_for?(work)
-        work.creator.blank? || work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
+      def cant_mint_for(work)
+        state = workflow_state(work)
+        cannot_use_workflow_state = ["deposited", nil].exclude?(state&.workflow_state_name)
+
+        !(work.creator.blank? || work.doi.present? || work.visibility != "open" || cannot_use_workflow_state)
       end
 
       def workflow_state(work)
--- a/app/jobs/hyku_addons/reindex_model_job.rb
+++ b/app/jobs/hyku_addons/reindex_model_job.rb
@@ -8,7 +8,6 @@ module HykuAddons
   class ReindexModelJob < ApplicationJob
     rescue_from Hyrax::DOI::DataCiteClient::Error, Ldp::Gone, Ldp::HttpError, RSolr::Error::Http, RSolr::Error::ConnectionRefused do |exception|
       Rails.logger.debug exception.inspect
-      retry_job(wait: 5.minutes)
     end
 
     # rubocop:disable Metrics/MethodLength
@@ -35,7 +34,7 @@ module HykuAddons
     private
 
       def mint_doi(work)
-        return if can_mint_for?(work)
+        return if cant_mint_for?(work)
 
         Rails.logger.debug "=== about to mint doi for #{work.title} ==== "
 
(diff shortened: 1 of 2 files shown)


app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -44,8 +43,8 @@ def mint_doi(work)
         work.update(doi: [register_doi.identifier])
       end
 
-      def can_mint_for?(work)
-        work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
+      def cant_mint_for?(work)
```
thenapking, line 46: It is better to use positive names for methods returning truth values to avoid double negation.  Why not use !can_mint_for(work) and then define that as:
````
def can_mint_for(work)
  work.creator.present? || work.doi.blank? || work.visibility == "open" || workflow_state(work)&.workflow_state_name == "deposited"
end

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: Thanks just seeing this.

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: @thenapking which do you say is preferred? because you said positive names is preferred  for method returning truth  but proposed the proposed name **!cant_mint_for(work)** looks like  the negative version instead of the current positive sounding version. 

Did i get you wrong.

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: is an if check not to run if any of those is true, but this returns false:

         false unless work.creator.present? 
         
  While in **mint_doi** method we have 
  
       return if cant_mint_for?(work)

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: I can switch that to unless

Comment (thenapking) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: I would do this if you are short of time:
````
def can_mint_for(work)
  !(work.creator.blank? || work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited")
end
````
And then call `!can_mint`.  It would then be worth refactoring this code as soon as you do have time.

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: Thanks I will go with that with a small change instead of calling **!cant_mint** another negation, I will just switch **if** to **unless**

Comment (mankind) [reply app/jobs/hyku_addons/reindex_model_job.rb:46]: I don't know what the conditions really are here and it's difficult to know whether you are changing things to correct a typo or to clear up the logic.

For me the change will be to correct typo from cant to can.

File: app/jobs/hyku_addons/reindex_model_job.rb [hunk 2]
```diff
@@ -35,7 +34,7 @@ def perform(klass, cname, limit: 25, page: 1, options: {})
     private
 
       def mint_doi(work)
-        return if can_mint_for?(work)
+        return unless cant_mint_for(work)
```
thenapking, line 37: `unless cant` is a double negative and should be avoided.  The method is named negatively, and the whole condition at the end is negated.  This is even more confusing than before.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -44,8 +43,11 @@ def mint_doi(work)
         work.update(doi: [register_doi.identifier])
       end
 
-      def can_mint_for?(work)
-        work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
+      def cant_mint_for(work)
+        state = workflow_state(work)
+        cannot_use_workflow_state = ["deposited", nil].exclude?(state&.workflow_state_name)
+
+        !(work.creator.blank? || work.doi.present? || work.visibility != "open" || cannot_use_workflow_state)
```
thenapking, line 50: The method is named negatively, and this condition at the end is negated.  If this actually works please just name it positively, and remove the !.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -44,8 +43,11 @@ def mint_doi(work)
         work.update(doi: [register_doi.identifier])
       end
 
-      def can_mint_for?(work)
-        work.doi.present? || work.visibility != "open" || workflow_state(work)&.workflow_state_name != "deposited"
+      def cant_mint_for(work)
```
prdanelli, line 46: Could you use the `?` suffix for methods like this, so that we know its a boolean result being returned. 

Why are you using a negative and not a positive that is inverted? I think using negative methods makes it harder keep track of whats going on.

app/jobs/hyku_addons/reindex_model_job.rb
mankind: Thanks we might need to pack the pr or may be revert to the original since this current changes was to try and address what you said was confusing. 

I will revert to the original and push and it is still confusing, I am happy to throw away the PR and leave the job as it is for now and return to it later.

prdanelli (commented):
I just have a question about the negative method naming.

app/jobs/hyku_addons/reindex_model_job.rb
thenapking: This is entirely fixable.  Please do this:
````
def can_mint_for?(work)
  -- conditions under which a work can be minted 
end
```

Then call it in the way you want to test for... if something can be minted, or unless it can be minted.  I don't understand why this is becoming so difficult.  Please do not revert because it is now unclear in your original code whether you had the method named wrongly all along.

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -44,8 +43,11 @@ def mint_doi(work)
...
+        state = workflow_state(work)
+        cannot_use_workflow_state = ["deposited", nil].exclude?(state&.workflow_state_name)
```
thenapking, line 48: Also you have defined another negative here.  Try instead:
````
can_use_workflow_state = ["deposited", nil].include?(state&.workflow_state_name)
````

Then instead call `!can_use_workflow_state`.

mankind: Thanks PR being closed for such a time as the job is needed again.

prdanelli:
@mankind would it be useful to merge the specs into the `main` branch?

mankind:
> @mankind would it be useful to merge the specs into the `main` branch?

No need mate, the PR is closed. Thanks ):

PR #612, opened 2022-01-19
Allow Bulkrax to import collection titles

thenapking:
Override `collections` method in `CsvParser` so it retrieves the collection title from the CSV file.
Refactor `create_collections` to move shared logic to a new method
Future proof this against changes in Bulkrax v3 which will remove `collection` column and replace it with `parent` column
Add tests


app/parsers/concerns/hyku_addons/collection_behavior.rb
```diff
@@ -0,0 +1,100 @@
...
+          visibility: 'open',
+          collection_type_gid: Hyrax::CollectionType.find_or_create_default_collection_type.gid
+        }
+      end
+
+      def collection_delimiter
+        Bulkrax.field_mappings["HykuAddons::CsvParser"]&.dig("collection", :split) || "\|"
+      end
+
+      def collection_prefix
```
mankind, line 74: This not asking for anything but raising it for discussion. We are still on **bulkrax (0.1.0 82dffb2)** while the current Bulkax version is 2. Is the work a good opportunity to upgrade our bulkrax to version 2 and we could do remove this code.
The only other change after the upgrade will be for Tom M to update the csv files they use to use **parent** and **child** instead of the current collection mapping.

app/parsers/concerns/hyku_addons/collection_behavior.rb
```diff
@@ -0,0 +1,100 @@
...
+            { id: collection_ids(r)[i], title: collection_titles(r)[i] }
+          else
+            { id: id, title: "New collection #{i + 1}" }
+          end
+        end
+      end.flatten.compact.uniq
+    end
+
+    # Override to pass Bulkrax.system_identifier_field as singular instead of array
+    def create_collections
```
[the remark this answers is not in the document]
thenapking: Since both loops were very similar, I instead took the path of extracting them both into one method, rather than two.

app/parsers/concerns/hyku_addons/collection_behavior.rb
thenapking: I haven't added anything to test parent or child columns *in general* as that should be done when you update to Bulkrax.  All this does is test whether the particular parent column will work.  I think that updating to Bulkrax 3 should probably be raised as a separate ticket, so we can assess the priority of that.  This could be a good time to do it, I don't know.

mankind: Thanks, I will ask other dev during Standup and then perhaps Elisa & Tom later since they might not know how far behind our version is, so they can decide since this feature is actively used.

PR #617, opened 2022-01-21
Skip retry in reindexmodeljob

mankind:
During minting of doi when exceptions are raised, the work is re-queued. The exception will still be handled but without retries.
Also if a work is sent for doi minting and it has no creator as was the case for some imported works, the minting will fail but in Datacite it will create a draft doi that is not attached to any work. We want to avoid that, by not minting doi for works that have no creator.


thenapking (commented):
This was showing code climate issues which I think relate to old code.  I have rebased to main and force pushed to clear them.

thenapking (approved):
LGTM

PR #620, opened 2022-01-21
Port ImportWorkCollectionJob from old version of Bulkrax

thenapking:
Port old version of ImportWorkCollectionJob from Bulkrax and call it directly from CollectionBehavior.
During testing on AH it became apparent that repeatedly running the same import causes odd errors to arise and causes collection titles to be overwritten with IDs.  This appears to solve those issues locally, however if it works we would still need to know why...

### DIFF
--- a/app/jobs/hyku_addons/import_work_collection_job.rb
+++ b/app/jobs/hyku_addons/import_work_collection_job.rb
@@ -1,0 +1,41 @@
+# frozen_string_literal: true
+
+module HykuAddons
+  class ImportWorkCollectionJob < ApplicationJob
+    queue_as :import
+
+    # rubocop:disable Rails/SkipsModelValidations
+    def perform(*args)
+      entry = Bulkrax::Entry.find(args[0])
+      begin
+        entry.build
+        entry.save
+        add_user_to_permission_template!(entry)
+        Bulkrax::ImporterRun.find(args[1]).increment!(:processed_collections)
+        Bulkrax::ImporterRun.find(args[1]).decrement!(:enqueued_records)
+      rescue => e
+        Bulkrax::ImporterRun.find(args[1]).increment!(:failed_collections)
+        Bulkrax::ImporterRun.find(args[1]).decrement!(:enqueued_records)
+        raise e
+      end
+    end
+    # rubocop:enable Rails/SkipsModelValidations
+
+    private
+
+      def add_user_to_permission_template!(entry)
+        user                = ::User.find(entry.importerexporter.user_id)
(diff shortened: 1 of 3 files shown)


BertZZ (approved):

mankind:
I think what was copied over is an exact match with the content of import_work_collection_job.rb  file in our version of Bulkrax that is 0.1.0.

I did a blame on  Github Bulkrax repo and went to just before version 1 and it showed that the content of import_work_collection_job.rb was last changed 11months ago and we pinned our version about 6 -  7 months ago. 

Also in my local docker i did **bundle show bulkrax** then used **cat** to inspect the content of that file and it looks thesame with what was copied across in this **pr**. So hopefully my eyes  did not skip-over anything.

Kindly double check what I have observed before merging. Thanks

thenapking:
> I think what was copied over is an exact match with the content of import_work_collection_job.rb file in our version of Bulkrax that is 0.1.0.
> 
> I did a blame on Github Bulkrax repo and went to just before version 1 and it showed that the content of import_work_collection_job.rb was last changed 11months ago and we pinned our version about 6 - 7 months ago.
> 
> Also in my local docker i did **bundle show bulkrax** then used **cat** to inspect the content of that file and it looks thesame with what was copied across in this **pr**. So hopefully my eyes did not skip-over anything.
> 
> Kindly double check what I have observed before merging. Thanks

You are exactly right.  This is a direct port of the 0.1 code.  What is mystifying is that the errors when you re-run the same job multiple times disappear when you port this code in.  I don't think this is a good solution because it makes no sense why this should work.  But it might get us some way further into investigating the problem I can still observe with IDs being assigned to titles of collections.

mankind:
> > I think what was copied over is an exact match with the content of import_work_collection_job.rb file in our version of Bulkrax that is 0.1.0.
> > I did a blame on Github Bulkrax repo and went to just before version 1 and it showed that the content of import_work_collection_job.rb was last changed 11months ago and we pinned our version about 6 - 7 months ago.
> > Also in my local docker i did **bundle show bulkrax** then used **cat** to inspect the content of that file and it looks thesame with what was copied across in this **pr**. So hopefully my eyes did not skip-over anything.
> > Kindly double check what I have observed before merging. Thanks
> 
> You are exactly right. This is a direct port of the 0.1 code. What is mystifying is that the errors when you re-run the same job multiple times disappear when you port this code in. I don't think this is a good solution because it makes no sense why this should work. But it might get us some way further into investigating the problem I can still observe with IDs being assigned to titles of collections.

Thanks for the clarification :+1:

PR #624, opened 2022-01-25
Add Docker runner container

prdanelli:
Add a "runner" to the docker compose setup. 

Currently the main rails container needs to be started for you to perform any action on it, `bundle install`, `rspec`, `rails console`, which is annoying to say the least. With the runner, you can (after building the containers again with `docker-compose build` start a temporary container that will be removed after its used. 

```bash
docker-compose run --rm runner <command>

# You could (as I have) add an bash alias / fish function to run this command when the following is entered: 
dcr <command>

# In fish it is 
function dcr
	docker-compose run --rm runner $argv
end
```

This means that you could use `dcr bunde exec rails c` to run the console without needing to have the main `web` container running.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,6 +62,9 @@ FROM dependencies_image as development_image
 
 WORKDIR /home/app
 
+# Add the binstubs to the path to avoid requiring bundle exec to be prefixed
+ENV PATH=/home/app/bin:$PATH
+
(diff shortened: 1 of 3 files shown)


mankind:
> 

Can we not achieve same with:

         docker-compose run web bundle exec rails console

thenapking (approved):
LGTM

PR #627, opened 2022-01-26
Schema source location behavior

prdanelli:
Ensure that Schema defined models include the `source_location` field, which is required by Bulkrax. I have also performed a little boyscouting and moved the NotesBehavior to a shared concern for all schema models.

### DIFF
--- a/lib/hyku_addons/engine.rb
+++ b/lib/hyku_addons/engine.rb
@@ -273,13 +273,13 @@ module HykuAddons
         end
 
         def find
-          return find_by_id if attributes[:id]
-
-          if attributes[system_identifier_field].present? && klass.new.respond_to?(system_identifier_field)
-            return search_by_identifier
+          if attributes[:id]
+            find_by_id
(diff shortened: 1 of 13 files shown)


lib/hyku_addons/engine.rb
```diff
@@ -270,9 +273,13 @@ def update_admin_set(attrs)
...
-          return find_by_id if attributes[:id]
-          return search_by_identifier if attributes[system_identifier_field].present?
-          return search_by_title_or_identifier if klass == AdminSet && attributes[:title].present?
+          if attributes[:id]
+            find_by_id
+          elsif attributes[system_identifier_field].present? && klass.new.respond_to?(system_identifier_field)
+            search_by_identifier
+          elsif klass == AdminSet && attributes[:title].present?
+            search_by_title_or_identifier
+          end
```
prdanelli, line 282: Thanks for this @thenapking, its so much easier to read now .

PR #628, opened 2022-01-28
Schema JSON fields config for bulkrax importers

prdanelli:
Fix issue where schema json fields are not the same format as bulkrax importers expect


thenapking (approved):
LGTM.

thenapking (approved):
LGTM

PR #631, opened 2022-01-31
Account for multiple subfields in json blocks

prdanelli:
It doesn't seem like the multiple JSON field `funder_awards` was being transformed into an array if required. It was initially reported on UvaWork however, I am unsure if this was working elsewhere, but it doesn't appear so from my tests, or the code.


app/models/hyku_addons/importer_overrides.rb
```diff
@@ -76,10 +80,16 @@ def add_json_fields
...
-        raw_metadata.select { |k, _v| k.starts_with? field.to_s }.each do |k, v|
-          match = k.match(/^(?<subfield>.+)_(?<index>\d+)$/)
-          field_json[match[:index].to_i - 1] ||= {}
-          field_json[match[:index].to_i - 1][match[:subfield]] = v
+        raw_metadata.select { |key, _v| key.starts_with? field.to_s }.each do |key, value|
+          match = key.match(/^(?<subfield>.+)_(?<index>\d+)$/)
+
+          # If the value is defined as a multiple value and has a split delimiter defined in the field_mappings
+          # defined in the HykuAddons::Engine. i.e. "funder_award" => { split: '\|' },
+          if (delimiter = Bulkrax.field_mappings.dig("HykuAddons::CsvParser", match[:subfield], :split)).present?
```
thenapking, line 88: Could you make delimiter a separate method please so this is easier to read?

prdanelli: I think you're right. I tend to use these `(variabe = method.call()).present?` blocks, but when its just a single line it does look untidy.

PR #632, opened 2022-02-01
Ensure that the job does not find filesets

BertZZ:
The original job was finding filesets that do not have the creator field as JSON 

` #<FileSet id: "452f45c4-d849-4fa4-ad0b-5bc10c966a2b", head: [], tail: [], depositor: "bertie.wooles@ubiquitypress.com", title: ["Yoohyeon.jpg"], date_uploaded: "2021-10-18 10:54:48", date_modified: "2021-10-18 10:54:48", label: "Yoohyeon.jpg", relative_path: nil, import_url: nil, resource_type: [], creator: ["bertie.wooles@ubiquitypress.com"]`

Therefore trying to parse the creator of the filesets was causing the error, this change ensures only Works are returned


prdanelli (approved):
Good work finding this.

PR #633, opened 2022-02-01
Stop filesets showing up in users works

BertZZ:
Same issue as #632 but the creator is being parsed by the jbuilder file


prdanelli (changes requested):
As you're adding this again, we should extract it to a helper

BertZZ:
Yeah ok, I want to get the endpoint working so Rees is unblocked but I will look at a refactor next

prdanelli (approved):
Approved on the condition we come back to this next to DRY is up :)

PR #637, opened 2022-02-03
CSV for importer is not successfully importing publication dates

thenapking:
Remove CsvMatcher class which is redundant
Create a dynamic test for Bulkrax importer


spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    def get_expectations_for_field(attribute_name, attribute_properties, row, work)
+      attribute_test_value = attribute_properties[:multiple] ? [row[attribute_name]] : row[attribute_name]
+      attribute_actual_value = work.send(attribute_name)
+
+      { attribute_name: attribute_name, actual: attribute_actual_value, test: attribute_test_value }
+    end
+
+    def get_expectations_for_subfield(attribute_name, subfield, row, work)
+      attribute_test_value = row["#{subfield}_1"]
+      attribute_actual_value = JSON.parse(work.send(attribute_name).first, symbolize_names: true).first[subfield]
```
thenapking, line 169: Could someone comment on whether there is a better way to do this rather than parsing this field directly.

spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+    end
+
+    def untested_attributes
+      %i[resource_type language rights_statement relation_type subject display_creator_profile]
+    end
+
+    # CSV Methods
+
+    def headers
+      subfields_with_suffixes = subfields.transform_keys { |k| "#{k}_1".to_sym }
```
thenapking, line 89: This test does not yet add multiple creators etc.  I can refactor this method to do that if it is essential.

spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+                   .reduce({}, :merge)
+                   .reject { |k, _v| untested_attributes.include?(k) }
+    end
+
+    def attributes
+      field_configs_without_subfields.merge!(subfields)
+    end
+
+    def untested_attributes
+      %i[resource_type language rights_statement relation_type subject display_creator_profile]
```
thenapking, line 83: Please comment on whether all of these should be excluded.  The intention here is only to exclude internal fields which cause specific behaviours of the importer or application which should be tested in isolation, eg embargoing, DOI minting. I am unsure for instance why `display_creator_profile` never succeeds.

spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+            expectations << get_expectations_for_field(attribute_name, attribute_properties, row, work)
+          end
+        end
+      end
+
+      expectations
+    end
+
+    def get_expectations_for_field(attribute_name, attribute_properties, row, work)
+      attribute_test_value = attribute_properties[:multiple] ? [row[attribute_name]] : row[attribute_name]
```
thenapking, line 161: At the moment, I do not test adding multiple values for simplicity.

spec/features/bulkrax_dynamic_import_spec.rb
prdanelli: @BertZZ might know more about `display_creator_profile`. I thought it was part of the JSON fields.

prdanelli (dismissed):
I've got nothing to add, this is great. Thank you.

spec/features/bulkrax_dynamic_import_spec.rb
thenapking: I will rebase this on main and see if we can remove this once main is passing again.

spec/features/bulkrax_dynamic_import_spec.rb
thenapking: I have now added this to test multiple fields

spec/features/bulkrax_dynamic_import_spec.rb
thenapking: `display_creator_profile` is now no longer required! So we've definitely fixed this one.

spec/support/bulkrax/csv_reader_helper.rb
```diff
@@ -0,0 +1,52 @@
...
+    attribute_properties = attribute.last
+
+    if work.respond_to?(attribute_name)
+      if attribute_properties.dig(:subfields)
+        attribute_properties[:subfields].keys.reject { |k| untested_attributes.include?(k) }.each do |subfield|
+          expectations << get_expectations_for_subfield(attribute_name, subfield, attribute_properties[:multiple], row, work)
+        end
+      else
+        # TODO: Bulkrax sets rights_statement as a multiple field, but it is not
+        wrongly_configured_fields = ["rights_statement"]
```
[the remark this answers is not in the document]
thenapking: There's already a [ticket](https://ubiquitypress.atlassian.net/browse/REPO-1471) for it.

spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+  include CsvReaderHelper
+
+  let(:user) { create(:user, email: "test@example.com") }
+  let(:depositor) { build_stubbed(:user, email: "batchuser@example.com") }
+
+  let(:number_of_records) { 3 } # Setting this too high will increase the test time for little gain
+
+  after do
+    ["UbiquityTemplateWork", "UvaWork"].each do |model_name|
+      File.delete("spec/fixtures/csv/#{model_name.underscore}_dynamic_data.csv") if File.exist?("spec/fixtures/csv/#{model_name.underscore}_dynamic_data.csv")
```
prdanelli, line 18: There are a few really long lines of code in here that it would be great if we could get down to <= 120.

spec/features/bulkrax_dynamic_import_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        perform_enqueued_jobs(only: Bulkrax::ImporterJob) do
+          importer.import_collections
+        end
+
+        perform_enqueued_jobs(only: Bulkrax::ImporterJob) do
+          importer.import_works
+        end
+      end.to change { @model_name.constantize.count }.by(number_of_records)
+
+      aggregate_failures do
```
prdanelli, line 54: TIL.

prdanelli: This could be useful throughout the entire test suite, especially for bigger tests like the schema feature specs.

thenapking: I believe there's a rubocop setting which enforces using `aggregate_failures` so that all your tests are unit tests more or less.  You can use it on the describe block too.

prdanelli (approved):

PR #638, opened 2022-02-04
Refactors author profile code to keep name in line with convention and remove boolean

BertZZ:


app/actors/hyku_addons/actors/creator_institutional_email_actor.rb
```diff
@@ -18,11 +18,11 @@ def check_creator_institutional_email(env)
           return [] if creator_hash.blank?
 
           creators = creator_hash.map do |creator|
-            if %w[Organizational Organisational].include?(creator["creator_name_type"])
-            else
+            unless %w[Organizational Organisational].include?(creator["creator_name_type"])
               user = User.find_by(email: creator["creator_institutional_email"])
-              creator["display_creator_profile"] = user.present? && user.display_profile
+              creator["creator_profile_visibility"] = user.present? && user.display_profile ? "open" : "closed"
```
prdanelli, line 23: Its best to avoid having configuration strings distributed around the codebase like this. The optimal thing to do is keep all values like this on the User its self, so that i can be referred to globally and always return the same value and type.  

```ruby
module UserBehavior
#... 
    def display_profile_visibility
        user.display_profile ? "open" : "closed"
    end
end
```

Then you can reter to this value everywhere in the codebase without repeating strings

```ruby
creator["creator_profile_visibility"] = user.display_profile_visibility
```

As this change doesnt account for the user being present, you'll need an extra guard clause in your block that returns early if the user isn't found on the system: 

```ruby
# Instead of 
user = User.find_by(email: creator["creator_institutional_email"])

# You could do
return if (user = User.find_by(email: creator["creator_institutional_email"])).blank?
```

BertZZ: I see what you are saying now, Ok :+1:

app/models/concerns/hyku_addons/user_behavior.rb
```diff
@@ -21,10 +21,14 @@ def current_account
...
 
+    def display_profile_visibility
+      user.display_profile ? "open" : "closed"
+    end
+
     def toggle_display_profile
       return unless display_profile_changed?
       yield
-      HykuAddons::ToggleDisplayProfileJob.perform_later(email, display_profile)
+      HykuAddons::ToggleDisplayProfileJob.perform_later(email, display_profile_visibility)
```
prdanelli, line 31: Do you need to change this? `display_profile_visibility` is the string used in the JSON, rather than the users boolean preference.

BertZZ: Yeah because the Job changes the JSON string in all the works, otherwise you are doing the conditional again

prdanelli: I see, so the job updates the JSON to whatever value is passed in here. Cool :)

PR #639, opened 2022-02-04
Override Bolognese method that is incompatible with JSON field creators

prdanelli:


app/services/bolognese/writers/hyrax_work_writer_behavior.rb
```diff
@@ -29,6 +29,33 @@ def pluralize_and_process(type)
...
+
+          # remove spaces around hyphens
+          author = author.gsub(" - ", "-")
+
+          # titleize strings
+          # remove non-standard space characters
+          author.my_titleize.gsub(/[[:space:]]/, ' ')
+        end
+
+        def json?(author)
```
thenapking, line 53: Should this be a private method?

app/services/bolognese/writers/hyrax_work_writer_behavior.rb
```diff
@@ -29,6 +29,34 @@ def pluralize_and_process(type)
...
+
+          # detect pattern "Smith J.", but not "Smith, John K."
+          author = author.gsub(/[[:space:]]([A-Z]\.)?(-?[A-Z]\.)$/, ', \1\2') unless author.include?(",")
+
+          # remove spaces around hyphens
+          author = author.gsub(" - ", "-")
+
+          # titleize strings
+          # remove non-standard space characters
+          author.my_titleize.gsub(/[[:space:]]/, ' ')
```
thenapking, line 50: I can't see the code for `my_titleize` anywhere, could you link me to it?

spec/services/bolognese/writers/hyrax_work_writer_behavior_spec.rb
```diff
@@ -98,4 +98,40 @@
       end
     end
   end
+
+  describe "#cleanup_author" do
```
thenapking, line 102: I think we might be missing a test where one or both of the author's names are blank, but not nil. Can that happen?

PR #640, opened 2022-02-04
Remove the next as it was causing authors to be skipped

prdanelli:


BertZZ:
@prdanelli We can refactor this on Monday but I need to try start this job so I am going to merge

PR #641, opened 2022-02-07
Catch Solr connection error for test tenants that weren't setup correctly

prdanelli:


lib/tasks/task_master.rake
```diff
@@ -7,6 +8,17 @@ namespace :task_master do
     end
   end
 
+  task count_all: :environment do
+    Account.all.each do |account|
```
BertZZ, line 12: Do we also want to take into account @mankind Suggestion of skipping search tenants 
```suggestion
    Account.where(search_only: false).each do |account|
```

BertZZ, line 12: Do we also want to take into account @mankind Suggestion of skipping search tenants 
```suggestion
    Account.where(search_only: false).each do |account|
```

prdanelli: 100%. I wasn't aware of that switch. Can you add it to the PR?

PR #643, opened 2022-02-07
Refactor User Profile Visibility

prdanelli:
Make some small changes to the structure and naming.


app/actors/hyku_addons/actors/creator_profile_visibility_actor.rb
```diff
@@ -0,0 +1,43 @@
+# frozen_string_literal: true
+module HykuAddons
+  module Actors
+    class CreatorProfileVisibilityActor < Hyrax::Actors::AbstractActor
```
prdanelli, line 4: @BertZZ I thought it would be a good idea to rename this class to better illustrate what it actually related to, rather than a name that was related to the creator field that it edited.

app/models/concerns/hyku_addons/user_behavior.rb
```diff
@@ -4,8 +4,11 @@ module HykuAddons
   module UserBehavior
     extend ActiveSupport::Concern
 
+    PROFILE_VISIBILITY = { open: "open", closed: "closed" }.freeze
```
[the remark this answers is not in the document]
BertZZ: What does this return on the API?

lib/hyku_addons/engine.rb
```diff
@@ -534,7 +535,10 @@ def self.dynamically_include_mixins
       Hyrax::CurationConcern.actor_factory.insert_before Hyrax::Actors::ModelActor, HykuAddons::Actors::DateFieldsActor
       Hyrax::CurationConcern.actor_factory.insert_before Hyrax::Actors::ModelActor, HykuAddons::Actors::NoteFieldActor
       Hyrax::CurationConcern.actor_factory.insert_before Hyrax::Actors::ModelActor, HykuAddons::Actors::RelatedIdentifierActor
-      Hyrax::CurationConcern.actor_factory.insert_after HykuAddons::Actors::JSONFieldsActor, HykuAddons::Actors::CreatorInstitutionalEmailActor
+
+      actors = [HykuAddons::Actors::JSONFieldsActor, HykuAddons::Actors::CreatorProfileVisibilityActor]
+      Hyrax::CurationConcern.actor_factory.insert_after(*actors)
```
prdanelli, line 540: @BertZZ The previous line was really long, so this is a nice way of breaking that down into an array and passing that to the `insert_after` method, where the array of extracted into arguments.

BertZZ:
I am ok with all of this :smiley:

app/actors/hyku_addons/actors/creator_profile_visibility_actor.rb
```diff
@@ -0,0 +1,43 @@
...
+        def toggle_visibility(env)
+          creators = extract_creators(env)
+
+          return [] if creators.blank?
+
+          creators.map! do |creator|
+            if creator["creator_name_type"] == "Personal"
+              user = User.find_by(email: creator["creator_institutional_email"])
+
+          byebug
```
thenapking: Whoops.  byebug left here.

app/models/concerns/hyku_addons/user_behavior.rb
```diff
@@ -4,30 +4,40 @@ module HykuAddons
...
-      accepted_email_formats = current_account&.settings&.dig('email_format')
-      return unless accepted_email_formats.present?
+      validate :email_format
 
-      email_format = '@' + email.split('@')[-1]
-      errors.add(:email, "Email must contain #{accepted_email_formats.join(', ')}") unless accepted_email_formats.include? email_format
+      scope :with_public_profile, -> { where(display_profile: true) }
     end
 
+    # NOTE: Is this used anywhere?
```
thenapking, line 17: It looks like it was used on line 30, and has now been removed and replaced with `Site.account`.  But a search of the code base turns up results for this method so perhaps double check before you delete it.  But we might as well clear this up now.

app/models/concerns/hyku_addons/user_behavior.rb
```diff
@@ -4,30 +4,40 @@ module HykuAddons
...
+      PROFILE_VISIBILITY[display_profile ? :open : :closed]
+    end
+
+    protected
+
+    # Triggered when the user registers an account
+    def email_format
+      email_formats = Site.account&.settings&.dig("email_format")
+
+      return if email_formats.blank? || email_formats.include?("@#{email.split("@").last}")
```
thenapking: This seems a little obscure.  Can't we just use a regex to check the email is in the right format? This just checks it is an @ symbol followed by at least one character, which might not be alphanumeric.  We don't even know whether it has a domain. Are the email_formats just a list of domains the tenant's user's emails must belong to?

spec/actors/hyku_addons/actors/creator_display_profile_visibility_actor_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    stack.build(terminator)
+  end
+  let(:key) { "creator_profile_visibility" }
+
+  before do
+    allow(Site).to receive(:instance).and_return(site)
+  end
+
+  describe "#create" do
```
thenapking, line 41: There are no tests here where the email fails validation.  We need several scenarios where the email doesn't match the required format, is nil, blank or contains characters which are invalid.

spec/features/ubiquity_template_work_form_spec.rb
```diff
@@ -55,7 +55,7 @@
         creator_orcid: "0000-0000-1111-2222",
         creator_institutional_relationship: "Research associate",
         creator_isni: "56273930281",
-        creator_profile_visibility: "closed"
+        creator_profile_visibility: User::PROFILE_VISIBILITY[:closed]
```
thenapking, line 58: There are no tests here for an invalid form being submitted.  Is that tested elsewhere and if so, can we add the scenario when the email is invalid to that test?

spec/models/concerns/hyku_addons/user_behavior_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+    before do
+      user && public_user
+    end
+
+    it "doesn't return private users" do
+      expect(query).to include(public_user)
+      expect(query).not_to include(user)
+    end
+
+    it "doesn't affect normal queries" do
```
thenapking, line 19: I don't think this test is required.  It would only affect normal queries if `default_scope` was defined.  Is it defined somewhere?

spec/models/concerns/hyku_addons/user_behavior_spec.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
+RSpec.describe HykuAddons::UserBehavior, type: :model do
+  subject(:user) { create(:user) }
```
thenapking, line 4: Is it possible to use `build_stubbed`?

spec/models/concerns/hyku_addons/user_behavior_spec.rb
```diff
@@ -0,0 +1,75 @@
...
+
+    it "does not enque a job unless the display profile setting has been changed" do
+      user.department = "software development"
+
+      expect { user.save }.not_to have_enqueued_job(HykuAddons::ToggleDisplayProfileJob)
+    end
+  end
+
+  describe "#display_profile" do
+    it "defaults to false" do
```
thenapking, line 51: I'm presuming this means when you create a user with `display_profile =nil` then it gets set to false.  What happens if you try to update a user and set `display_profile = nil`?

app/models/concerns/hyku_addons/user_behavior.rb
prdanelli: @BertZZ you added this method, did you just use it in this class, or was it used elsewhere too?

spec/actors/hyku_addons/actors/creator_display_profile_visibility_actor_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+      {
+        creator_name_type: "Personal",
+        creator_family_name: "Hawking",
+        creator_given_name: "Stephen",
+        creator_institutional_email: user.email,
+        creator_profile_visibility: User::PROFILE_VISIBILITY[:closed]
+      }
+    ].to_json
+  end
+  let(:work) { create(:work, creator: [creator]) }
```
thenapking, line 17: Can we try using `build_stubbed` in this test too?

app/models/concerns/hyku_addons/user_behavior.rb
prdanelli: @thenapking I would like to have done this with the Rails validation `format`, so i'm seeing if thats possible, but i'm not sure it is given the error message and moving all this code into a lamba on the `validate` call doesn't seem better to me.

spec/models/concerns/hyku_addons/user_behavior_spec.rb
prdanelli: @thenapking honestly, i dont trust Active Fedora, so i wanted to be sure that this worked as I would expect.

spec/models/concerns/hyku_addons/user_behavior_spec.rb
prdanelli: @thenapking i dont think so, as its testing DB scopes.

spec/models/concerns/hyku_addons/user_behavior_spec.rb
prdanelli: @thenapking `display_profile` has a db level default of `false` on a boolean column, so i dont think it can be `nil`.

app/models/concerns/hyku_addons/user_behavior.rb
thenapking: How do we check the validity of the email_formats? Could you link me to that piece of code? My feeling is that it would be better to store email_domains, not formats, then check that the email includes one of those domains when split at the @ symbol.  But email validation is always tricky and we seem to be rolling our own here for no reason.

spec/models/concerns/hyku_addons/user_behavior_spec.rb
thenapking: Great!

app/models/concerns/hyku_addons/user_behavior.rb
prdanelli: @thenapking normally i would agree that this could be tweaked, however at this stage, maintaining the functionality is my primary concern because we need to work out a migration for existing clients could be required. 

But yes, storing the `@` and the domain is excessive.

spec/actors/hyku_addons/actors/creator_display_profile_visibility_actor_spec.rb
thenapking: Sorry ignore this comment, a test for that should be in the user behavior spec

PR #645, opened 2022-02-08
Remove blanket rubocop file excludes

prdanelli:
HA used a huge .rubocop.yml file to basically exclude hundreds of files to any kind of a check. This is a really bad idea, not only is the first error not checked, but nothing else is. Please so not add anything to the .rubocop.yml configuration unless its absolutely necessary as adding a `disable` comment around the item is a much better idea.


app/services/hyku_addons/jwt_cookies_service.rb
```diff
@@ -14,8 +14,9 @@ def initialize(user, env)
       @request = env.request
     end
 
-    def set_jwt_cookies(user = nil)
+    def assign_jwt_cookies(user = nil)
```
thenapking, line 17: This renaming does clear up the code.  Did rubocop spot this because it doesn't like setters named with set_? If that's so how did the private method get passed?

hyku_addons.gemspec
```diff
@@ -4,7 +4,7 @@ $LOAD_PATH.push File.expand_path("lib", __dir__)
 # Maintain your gem's version:
 require "hyku_addons/version"
 
-# Describe your gem and declare its dependencies:
+# rubocop:disable Metrics/BlockLength
```
thenapking, line 7: I'd be happy with system files like this being left on the exclusion list.

thenapking (approved):
LGTM!

thenapking (approved):
Fixed code climate

app/actors/hyku_addons/actors/date_fields_actor.rb
```diff
@@ -22,6 +22,8 @@ def serialize_date_fields(env)
         end
 
         def transform_date(date_hash, field)
+          return date_hash unless date_hash.is_a? Hash
```
thenapking, line 25: Whoops that one was mine.

thenapking (approved):
LGTM

PR #647, opened 2022-02-09
Add override for creator_profile_visibility

thenapking:
Add an override to the attribute value faker to always set `creator_profile_visibility` to `closed`

### DIFF
--- a/spec/support/bulkrax/csv_writer_helper.rb
+++ b/spec/support/bulkrax/csv_writer_helper.rb
@@ -75,7 +75,14 @@ module CsvWriterHelper
   def faked_attribute(name, i)
     return unless name.is_a? Symbol
 
-    if attributes[name][:type] == "date"
+    if name == :creator_profile_visibility
+      # TODO: This is wrong
+      # It should be an authority field, unfortunately no authority has been defined
+      # You should also be able to use User::PROFILE_VISIBILITY.values.sample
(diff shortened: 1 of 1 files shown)


spec/support/bulkrax/csv_writer_helper.rb
```diff
@@ -75,7 +75,14 @@ def fake_row_data(i)
   def faked_attribute(name, i)
     return unless name.is_a? Symbol
 
-    if attributes[name][:type] == "date"
+    if name == :creator_profile_visibility
```
prdanelli, line 78: I do have one small suggestion for this. To try and avoid a scenario where by each time a field like this is added, this method needs to be adjusted, we could add something like... 

```ruby 
  def faked_attribute(name, i)
    return unless name.is_a? Symbol
    
    method_name = "fake_#{name}"
    if respond_to?(method_name)
        send(method_name)
    elsif attributes[name][:type] == "date"
    # ...
```

This way we have specifialist methods to deal with difficult fields/keys rather than an increasingly long if/case statement here. 

What do you think?

thenapking: I think that we should not have any special keys or exceptions here.  This is a temporary patch so that main passes again.  I think this list should only contain attribute types.  Had this test been in place when the creator_profile_visibility PR was being worked on it would have pointed to the fact that could have been done differently.

prdanelli (approved):

PR #651, opened 2022-02-10
Redirect user to profile after registration

prdanelli:


lib/hyku_addons/engine.rb
```diff
@@ -49,17 +49,26 @@ class Engine < ::Rails::Engine
...
         end
 
         def current_account
           Site.account
         end
+
+        protected
+
+          # Redirect the user to the profile after registration
+          def after_sign_up_path_for(resource)
```
BertZZ, line 69: Where is this used?

prdanelli: Its part of Devise. https://github.com/heartcombo/devise/wiki/How-To:-Redirect-to-a-specific-page-on-successful-sign-up-%28registration%29

lib/hyku_addons/engine.rb
```diff
@@ -49,17 +49,26 @@ class Engine < ::Rails::Engine
...
         end
 
         def create
           return super if current_account&.allow_signup == "true"
-          redirect_to root_path, alert: t(:'hyku.account.signup_disabled')
+
+          redirect_to root_path, alert: t("hyku.account.signup_disabled")
         end
 
         def current_account
```
thenapking, line 62: I think this method could be better named because it is confusing in the context of what is usually called account sign-up.  Also could it be a private method?

spec/controllers/hyku/registrations_controller_spec.rb
```diff
@@ -0,0 +1,35 @@
...
+  let(:account) { create(:account) }
+  let(:site) { Site.new(account: account) }
+
+  before do
+    allow(Site).to receive(:instance).and_return(site)
+
+    @request.env["devise.mapping"] = Devise.mappings[:user] # rubocop:disable RSpec/InstanceVariable
+  end
+
+  context "with account signup enabled" do
```
thenapking, line 13: There should be a context where account signup is disabled and we can't create accounts.

prdanelli: @thenapking Those scenarios are covered inside of Hyku :)

lib/hyku_addons/engine.rb
prdanelli: So @BertZZ added this for testing I think, but i'm not 100% sure. We can rename it, but I would prefer not to do it in this PR.

spec/controllers/hyku/registrations_controller_spec.rb
prdanelli: @thenapking Those scenarios are covered inside of Hyku :)

thenapking: Oh year, I see that bit did not change

lib/hyku_addons/engine.rb
thenapking: Oh yes, also not your change. I must pay attention better.

thenapking (approved):
LGTM

PR #654, opened 2022-02-11
Changes location of hyku addons add ons

BertZZ:


BertZZ:
Unneeded as redirects to samvera-labs are automatic

PR #660, opened 2022-02-18
Migrate Anschutz Work to Schema YAML

prdanelli:
Convert the AnshutzWork type to use the Schema YAML. This migration was a little more complicated because of changes to the DOI elsewhere, that had a number of knock on effects here. For other migrations, the process should be much simpler than this PR would have you think. 

This PR is ready to be merged and tested in a production environment.


mankind (dismissed):
:+1:

thenapking (dismissed):
LGTM.  This will be a great when it is applied to the other work types.

PR #663, opened 2022-02-21
Move DOI field into schema now it is required in main form fields section

prdanelli:


config/metadata/partials/doi.yaml
```diff
@@ -0,0 +1,13 @@
+---
+doi:
+  type: string
+  predicate: http://purl.org/ontology/bibo/doi
+  multiple: true
```
thenapking, line 5: Why is multiple `true` here? I thought it was being set false everywhere?

prdanelli: Good spot, thank you. I had forgotten to update the default YAML

thenapking (approved):
LGTM

PR #668, opened 2022-02-24
Fix Collection Banner url

thenapking:
Replace original code


BertZZ:
Ahhhh that bit of code. Nice catch

PR #669, opened 2022-02-24
fix hyrax-autopopulation not adding modules to ancestors of hyku-workbase and shema work base

mankind:


prdanelli:
Superseded  by https://github.com/ubiquitypress/hyku_addons/pull/673

PR #671, opened 2022-02-28
Allow Bulkrax to assign queue based on number of rows in CSV

thenapking:


app/jobs/concerns/hyku_addons/portable_active_fedora_behavior.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+# rubocop:disable Rails/Output
```
thenapking, line 2: This disable statement should be removed, along with the puts statement before merge

app/jobs/concerns/hyku_addons/portable_bulkrax_entry_behavior.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+# rubocop:disable Rails/Output
```
thenapking, line 2: This disable statement should be removed, along with the puts statement before merge

app/jobs/concerns/hyku_addons/portable_generic_behavior.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+# rubocop:disable Rails/Output
```
thenapking, line 2: This disable statement should be removed, along with the puts statement before merge

app/jobs/hyku_addons/reindex_model_job.rb
```diff
@@ -21,18 +19,29 @@ def perform(klass, cname, limit: 25, page: 1, options: {})
...
+        fetch_work_using_ids
       else
-        reindex_and_mint_work(work_class, offset)
+        reindex_and_mint_work
       end
     end
 
     private
 
+      def works
```
thenapking, line 33: These changes clean up the class a little and allow the new code to share logic with the existing code.  However they are not strictly needed for this update.  Only the `total_entries` method is needed and that could be refactored if you do not wish to add all these refactors.

PR #673, opened 2022-02-28
update hyrax-autopopulation to include commit that adds the doi of works that are not open access to rejected_doi_list when autopopulating works

mankind:
Also contains code using **RSolr.escape** to escape special characters eg **( )** in doi before queryng against Solr to avoid Solr Bad Request error.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -56,7 +56,7 @@ GIT
 
 GIT
   remote: https://github.com/ubiquitypress/hyrax-autopopulation.git
-  revision: cdce28269e563ef638a094ee1bce5109f79c13d5
+  revision: 341809b697bacd52309d70ad9bc451533eccd881
   branch: main
   specs:
     hyrax-autopopulation (0.1.0)


prdanelli (approved):

prdanelli:
Message from @mankind regarding this PR: 

> The two main changes the PR added in Hyrax-Autopopulation and we want to pull into HykuAddons are:
> 
> 1. It adds the doi of works that are not open access to the rejected_doi_list when auto-populating works. Before this PR  the only way to reject a doi is by checking the box in the UI and clicking the reject button which puts the doi of a work that is open access to the rejected doi and deletes that open access work. This PR puts the doi of non open access work to the reject list too behind the scene without admins doing anything.
> We used the rejected_doi_list to avoid auto-populating works with that doi if another person adds them again or if the admin tries to re-scan orcid again which will pull in all doi's in the various works in that orcid id again.
> 
> 2.  The PR also contains code using RSolr.escape to escape special characters e.g. '( )',  in doi before querying against Solr to avoid Solr Bad Request error.

PR #677, opened 2022-03-02
Support work locale in API services

BertZZ:


app/services/hyku_addons/qa_select_service.rb
```diff
@@ -41,8 +42,10 @@ def model_name
       end
 
       def tenant_locale
-        @tenant_locale ||= begin
-          return unless (account = Site.instance&.account)
+        @_tenant_locale ||= begin
+          return @locale.to_s.upcase if @locale.present?
+          account = Site.instance&.account
+          return unless account.present?
```
prdanelli, line 48: This method is getting a bit hectic. We have two guard statements and an inline OR assignment, so this is looking more and more like a code smell that we can clean this up and make it simpler. 

By adding a helper method for `account` and adding an `&.` we can really clean up the method. Then by moving both of the `.to_s.upcase` calls to the result of the assignment, we have a single line method that seem much easier to reason with. 

```ruby
      def tenant_locale
        @tenant_locale ||= (@locale || account&.settings&.dig("locale_name").presence || account&.name).to_s.upcase
      end

      def account
        Site.instance&.account
      end
```

prdanelli: Personally, I dont like all of the `&.`'s. It would be much nicer if `NilSite` returned a `NilAccount` that mimicked the `Account` class, rather than just returning `nil` and causing these kinds of affects.

prdanelli (changes requested):
It's almost there, but there are a couple of places that I think we should try and clean up a liltle.

app/services/hyku_addons/qa_select_service.rb
```diff
@@ -25,29 +26,31 @@ def initialize(authority_name, model: nil)
...
       end
 
       def model_authority_name
-        [@authority_name, model_name].compact.join("-")
+        [@authority_name, model_name].reject(&:blank?).join("-")
       end
 
       def model_tenant_authority_name
-        [@authority_name, model_name, tenant_locale].compact.join("-")
+        [@authority_name, model_name, tenant_locale].reject(&:blank?).join("-")
```
prdanelli, line 37: We're repeating `.reject(&:blank?).join("-")` 3 times in this class, so that would indicate that we can move that to a method. `clean` is just a suggested name. 

```ruby 
      def model_tenant_authority_name
        clean([@authority_name, model_name, tenant_locale])
      end

      def clean(array)
          array.reject(&:blank?).join("-")
      end
```

prdanelli (changes requested):
Almost there.

app/services/hyku_addons/qa_select_service.rb
prdanelli: `prepare_to_string`, or something that describes what it does without being super verbose.

prdanelli (approved):

PR #811, opened 2023-01-31
Repo 1955 workos

cococoder:
Implement WorkOS SSO authentication for HykuAddons module

This commit implements WorkOS Single Sign-On (SSO) authentication for the HykuAddons module. 
The `auth` method redirects users to the WorkOS authorization URL, and the `callback` method handles the response from WorkOS and signs the user in. The user's email, password, and JWT tokens are stored in the application.

To get the rubocop changes working I have to include it in the PR, there is another PR with the same changes.


cococoder:
Merged on to main in a different commit.

PR #819, opened 2023-02-07
Authenticate for existing users

cococoder:
This fix moves the sign_in and JWT token outside of the block.
Clean up the use of the keys in the tests. The ENV variables will have
to be added to the CI for the tests to work.


cococoder:
CI not running, checked tests and rubocop on local machine

